PR #9, opened 2023-06-04
Feature/1 bbc clean data extractor

frenchfaso:


bbc_clean/treerez/README.md
```diff
@@ -0,0 +1,26 @@
+# TreeRez
```
zgypa, line 1: Maybe you could add the reason why we needed to develop this tool in the first place.

bbc_clean/treerez/errors.txt
```diff
@@ -0,0 +1,6 @@
+ERROR:root:ERROR opening: F:/Bolton Backup/Bolton-Brush Study Files/Backup Broadbent Bolton Collection of Patient X-rays\B3582\Frontal\B3582F10.tif
```
zgypa, line 1: Remove if we no longer need this file.

bbc_clean/treerez/requirements.txt
```diff
@@ -0,0 +1,2 @@
+numpy==1.24.2
```
zgypa, line 1: Any reason to lock to specific versions for this simple tool?

bbc_clean/treerez/treerez.py
```diff
@@ -0,0 +1,116 @@
...
+def get_scans(patients_sample) -> list:
+	""" Return a list with all the scans of the sampled patients """
+	
+	scans = []
+	for patient in patients_sample:
+		xray_types = os.scandir(os.path.join(source_folder, patient))
+		for xray in xray_types:
+			if(xray.is_dir()):
+				for img in os.scandir(xray.path):
+					if(img.name.endswith(('tif','tiff'))):
```
zgypa, line 39: What about upper case extensions? Will this catch *.TIF files as well? If not just make it case insensitive.

zgypa (changes requested):

bbc_clean/treerez/requirements.txt
frenchfaso: Python module dependencies can be quite finicky... I would rather be explicit about versions we know for sure work happily together.

zgypa (approved):

PR #12, opened 2024-10-07
Feature/8 evaluation of splitting algorithms

frenchfaso:


documentation/.gitignore
zgypa, line 1: If the code needs a working directory, better assign one, like working/ or training/ or temp/. Otherwise you shouldn't need to ignore files like this.

zgypa (commented):
OK, see comment on .gitignore.

Issue #26, opened 2025-11-29
Implement DRF for Backend/Frontend Separation


frenchfaso:
### Goal
Refactor the application architecture to strictly separate backend logic from frontend presentation.

### Description
We are transitioning to an architecture where Django acts as an API provider rather than rendering HTML directly. This involves implementing Django Rest Framework (DRF) to expose existing data models via a RESTful API. This change ensures a clear separation of concerns, allowing the frontend to consume data via standard HTTP endpoints.

### Key Deliverables
- DRF installation and configuration.
- API endpoints (Serializers/ViewSets) for core models.
- CORS configuration for frontend communication.

Issue #22, opened 2025-10-12
Add necessary docker files to build and deploy docker image.


zgypa:
We need to have scripts to build the docker and deploy it at a CR, or directly on to wingate.

- build proper docker image with all django dependencies.

PR #28, opened 2025-12-01
Feature/26 implement drf for backendfrontend separation

### LINKED ISSUE #24: Add Automatic Flipping and Rotation and Classification from RGWizard
Take the AI model from RGWizard and implement it here, using an API. Ideally strutucture as orthovision ai 
https://medoco-health.github.io/orthovision-ai-ig/StructureDefinition-orthovision-ai-bundle.html

but lets do we can for the 10 december meeting.

frenchfaso:
## Summary
Implements the complete REST API backend and establishes a production-ready infrastructure.

## Key Changes
- **API:** Full DRF implementation with nested routers and file upload support.
- **Docs:** Integrated OpenAPI 3.0 (Swagger/Redoc) via `drf-spectacular`.
- **Infra:** Dockerized with Gunicorn and WhiteNoise for static files.
- **Security:** Implemented `.env` management and global authentication.

Closes #24 #26 #22


bfd9000_web/archive/management/commands/init_codings.py
zgypa: we discussed having coding initialization as a migration and using official dicom and snomed codes, like the ones defined in the spreadsheet. What is the reason for doing them like this, using custom codes? If there is one, it should be documented in the python header docs.

bfd9000_web/archive/constants.py
zgypa: if any, use bfd9000.open-ortho.org

.env.example
zgypa: you might want to rename to dot-env.example, to clarify and make it more visibile to the develoepr.

bfd9000_web/archive/migrations/0003_initial_codings.py
zgypa: ask MC about this, i tend to agree with this:

The migration will probably work in many environments, but it has a few fragility points and one important operational caveat (non-reversibility). I recommend a small rewrite to make it robust and reversible. Below I summarize the issues I found, the risk they pose, and a small, safe replacement migration you can apply instead.

Risks and issues (what could break, and when)
1) Top-level import of archive.constants (fragility)
   - Problem: importing app code from a migration module at the top level is fragile. When Django imports migrations it will import that module; if archive.constants later changes (or begins importing something that’s not available at migration time), or if someone runs migrations after changing code structure, import can fail. Migrations should be as self-contained and stable as possible.
   - Severity: Medium. Usually constants-only modules are stable, but it’s a brittle practice and can produce confusing failures during deploys / CI when code and migrations mismatch.

2) No reverse (migration is not reversible)
   - Problem: RunPython(init_codings) is provided without reverse_code. If you later try to migrate backwards (rollback), Django cannot automatically remove the inserted Codings. That may be acceptable, but you need to be explicit.
   - Severity: Medium-low (depends on your rollback policy). If you want clean reversibility, add a reverse function that deletes the created codings.

3) get_or_create with defaults does not update existing rows
   - Problem: if an existing Coding exists with the same (system, code) but a different display value, get_or_create returns the existing row and does not update display. If the intent was to ensure display text is current, use update_or_create or explicitly update after get_or_create.
   - Severity: Low (only relevant when display text must be updated by migration).

4) Depending on model fields being present
   - Problem: init_codings calls Coding.objects.get_or_create(..., defaults={'display': display}). This assumes the Coding model (as of the migration’s app state) has a 'display' field. If the model schema had not included 'display' until a later migration, this would fail. In your case the repository already has a Coding model with 'display' (normal), but it's good to be cautious.
   - Severity: Low (likely fine here since Coding is a core model; nonetheless keep migrations aligned to the historical model state).

5) Idempotency & repeated runs
   - get_or_create ensures the migration is idempotent in creation (won’t duplicate entries), but because display is placed in defaults only, it won’t enforce display text if you expected that.

6) Runtime vs migration-time constants
   - Since constants.py is new in the same PR, applying migrations in a fresh clone will have constants.py present (so import will succeed). But best practice remains to avoid importing app modules inside migration modules.

Suggested fixes
- Avoid importing archive.constants at module level. Instead embed the system strings inside the migration (literal strings) or define them inside the RunPython callable. That makes the migration self-contained.
- Use update_or_create rather than get_or_create if you want to ensure the display is set to the value in the migration.
- Provide a reverse function that removes the created codings (or at least documents that the migration is irreversible).
- Keep the migration atomic (RunPython is atomic when migrations support atomic; fine).
- If you want to be extra-safe, explicitly handle cases where Coding model fields might differ in older historical models (rare here).

Proposed safer migration
- Below is a replacement migration that:
  - defines system constants inside the migration function (no top-level import of app code),
  - uses update_or_create to ensure the display matches,
  - provides a reverse function to delete the inserted codings (so the migration can be reversed cleanly),
  - is idempotent.

```python name=bfd9000_web/archive/migrations/0003_initial_codings.py url=https://github.com/open-ortho/edu.case.BFD9000/blob/main/bfd9000_web/archive/migrations/0003_initial_codings.py
# Generated by Django 5.2.8 on 2025-12-02 20:03 (REWRITTEN FOR STABILITY)
from django.db import migrations

def init_codings(apps, schema_editor):
    Coding = apps.get_model('archive', 'Coding')

    # Define systems locally to avoid importing app modules at migration import time
    SYSTEM_RECORD_TYPE = 'http://snomed.info/sct'
    SYSTEM_ORIENTATION = 'http://snomed.info/sct'
    SYSTEM_MODALITY = 'http://dicom.nema.org/resources/ontology/DCM'
    SYSTEM_PROCEDURE = 'http://snomed.info/sct'
    SYSTEM_BODY_SITE = 'http://snomed.info/sct'

    # Helper that uses update_or_create to ensure display is set/updated
    def ensure(system, code, display):
        Coding.objects.update_or_create(
            system=system,
            code=code,
            defaults={'display': display}
        )

    # Record Types
    record_types = [
        ('201456002', 'Cephalogram'),
        ('268425006', 'Pelvis X-ray'),
        ('39714003', 'Skeletal X-ray of wrist and hand'),
        ('1597004', 'Skeletal X-ray of ankle and foot'),
        ('302189007', 'Dental model'),
    ]
    for code, display in record_types:
        ensure(SYSTEM_RECORD_TYPE, code, display)

    # Orientations (projections & laterality)
    orientations = [
        ('399198007', 'Right lateral projection'),
        ('399173006', 'Left lateral projection'),
        ('272479007', 'Posteroanterior projection'),
        ('399348003', 'Antero-posterior projection'),
        ('7771000', 'Left'),
        ('24028007', 'Right'),
    ]
    for code, display in orientations:
        ensure(SYSTEM_ORIENTATION, code, display)

    # Modalities (DICOM)
    modalities = [
        ('RG', 'Radiographic Imaging'),
        ('SI', 'Scanned Image'),
        ('XC', 'External-camera Photography'),
        ('M3D', 'Model for 3D Manufacturing'),
    ]
    for code, display in modalities:
        ensure(SYSTEM_MODALITY, code, display)

    # Body Sites (SNOMED)
    body_sites = [
        ('69536005', 'Head'),
        ('609617007', 'Structure of pelvic segment of trunk'),
        ('731298009', 'Entire ankle and foot'),
        ('729875002', 'Entire wrist and hand'),
        ('1927002', 'Entire left elbow region'),
        ('71889004', 'Entire right elbow region'),
        ('210659002', 'Entire left knee'),
        ('210562007', 'Entire right knee'),
    ]
    for code, display in body_sites:
        ensure(SYSTEM_BODY_SITE, code, display)

    # Procedures (currently empty placeholder)
    procedures = []
    for code, display in procedures:
        ensure(SYSTEM_PROCEDURE, code, display)

def remove_codings(apps, schema_editor):
    Coding = apps.get_model('archive', 'Coding')

    SYSTEMS_AND_CODES = [
        ('http://snomed.info/sct', [
            '201456002','268425006','39714003','1597004','302189007',
            '399198007','399173006','272479007','399348003','7771000','24028007',
            '69536005','609617007','731298009','729875002','1927002','71889004','210659002','210562007'
        ]),
        ('http://dicom.nema.org/resources/ontology/DCM', [
            'RG','SI','XC','M3D'
        ]),
    ]
    for system, codes in SYSTEMS_AND_CODES:
        Coding.objects.filter(system=system, code__in=codes).delete()

class Migration(migrations.Migration):

    dependencies = [
        ('archive', '0002_imagingstudy_source_file'),
    ]

    operations = [
        migrations.RunPython(init_codings, reverse_code=remove_codings),
    ]
```

Short rationale for the replacement
- No top-level import of archive.constants: makes the migration a stable artifact independent of later code changes.
- update_or_create: ensures the display is set/updated to the value in the migration.
- reverse_code provided: enables reversing the migration cleanly if needed.
- Idempotent and clear about what it creates/deletes.

Other runtime notes (not migration-specific, but useful)
- The codebase now uses FileField.source_file on ImagingStudy and the views use source_file.path and open FileField with PIL. If you ever switch to a remote storage backend (S3, GCS), FileField.path may not be available. To support remote storage, the thumbnail/image views should use storage.open(...) and avoid relying on .path or os.path.exists(source_file.path). (This is a runtime compatibility issue, not a migration one.)
- The tests included in the PR create required Collection and Coding test rows, so test-suite will be OK. If you want a fresh DB to be bootstrapped (so a fresh install without running tests still has codings), the migration should create those entries — which it does (but see the import caveat above).

Conclusion
- The migration as authored will likely succeed in normal installs now (because the constants file exists in the same PR), but it is fragile (top-level import) and non-reversible. I recommend applying the safer replacement above (or equivalent changes): move constants into the migration (or reference literal strings), use update_or_create, and add a reverse_code. This will make the migration robust across versions and safer to roll back.

bfd9000_web/Dockerfile
```diff
@@ -0,0 +1,52 @@
...
+
+# Copy wheels from builder
+COPY --from=builder /app/wheels /wheels
+COPY --from=builder /app/requirements.txt .
+
+# Install packages
+RUN pip install --no-cache /wheels/*
+
+# Copy project code
+COPY . .
```
zgypa: [nitpick] can you double check that you are not putting in the docker image anything that is not needed for the docker to run? Are you sure we don´t have anything in there? If we don't, its fine.

bfd9000_web/archive/migrations/0003_initial_codings.py
```diff
@@ -0,0 +1,98 @@
...
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('archive', '0002_imagingstudy_source_file'),
+    ]
+
+    operations = [
+        migrations.RunPython(init_codings),
+    ]
```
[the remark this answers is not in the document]
aspiringLich: eh. Migration 4 should undo this anyway

bfd9000_web/README.md
```diff
@@ -13,22 +13,43 @@ Alternatively just install django with python.
...
+python manage.py test archive.tests.test_valuesets
+```
+
+Useful test options:
+- `--failfast` - Stop after first failure
+- `-v 2` - Verbose output
+- `--keepdb` - Keep test database between runs (faster)
+- `--parallel` - Run tests in parallel
+
+**Note**: Tests automatically clean up uploaded media files. Test images are stored in a temporary directory that is deleted after tests complete, so they won't clutter your `media/uploads/` directory.
```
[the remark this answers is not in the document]
aspiringLich: Eh, seemed fine?

zgypa (changes requested):
There are a lot of new and old comments left unanswered. Some still addressed to @frenchfaso. Now that you are taking over, you need to address those as if they were addressed to you. Please do the following:

- for each comment (human or non-human), ensure you answer with your decision: "done" or "will not do/doesn't apply" with explanation why.
- do not resolve human comments. Leave them unresolved and let the commenter/reviewer close them
- you can resolve LLM comments, after you answered it and applied modifications (if applicable)

PR #31, opened 2025-12-08
Feature/23 initial django UI new

aspiringLich:
Initial implementation of the django UI. Currently, all the data is hardcoded and static, but it should be easy enough to swap out for real data once the backend is merged. Note there's some difficulty with developing the frontend since you need to install daisyUI / tailwindCSS and I did not want to include that in the git repo. Also note `PaginatedTableManager` in `index_base.html` which cuts down a LOT of repeated javascript code. Currently, the individual records / subject view are not finished. We don't have any of the actual records data / the subject view is tightly integrated with scanning (creating new records).

Next steps:
 - [ ] Hack something together for the subject view and get scanning to work! Possibly with BFD9020 AI stuff as well.
 - [ ] Integrate backend and frontend together, replace static data with API calls
 - [ ] Finish individual records / subject view, fully populated with all relevant data

### DIFF
--- a/bfd9000_web/archive/static/css/output.css
+++ b/bfd9000_web/archive/static/css/output.css
@@ -1429,6 +1429,191 @@
   .right-2 {
     right: calc(var(--spacing) * 2);
   }
+  .file-input {
+    @layer daisyui.l1.l2.l3 {
+      cursor: pointer;
+      cursor: pointer;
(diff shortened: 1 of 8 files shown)


zgypa:
Closing in favor of #38. The feature/33 branch contains all commits from #31 plus the backend integration and scan workflow. Please review #38 instead.

Issue #65, opened 2026-03-02
AI doesn't automatically select Radiographic Image


zgypa:
When a Cephalogram is detected, Radiographic Iimage should be automatically selected.

Related to #64

zgypa:
This field should probably not show, and it's confusing. We should only give the user ability to select the different kinds of records, which would be a local set of codes.

PR #70, opened 2026-03-02
Cleanup Scanning Page

### LINKED ISSUE #64: Missing Frontal (PA) type
When scanning, there is no PA frontal/coronal view. Lateral would be Sagittal.

aspiringLich:
Fixes #64 and #65

Instead of selecting the record type and modality, you select the image type and it figures it out. Currently hardcoded it in the submit function on the client side code.

There are some inconsistencies there—I'll make an issue about it.

`aiImageType` was redundant, `aiPatientOrientation` was not actually given by the AI. Although it is a field on the `Record` model it was pointless to set it.


bfd9000_web/archive/templates/archive/scan.html
```diff
@@ -1046,17 +1001,17 @@ <h3 class="font-bold text-lg mb-2">Scanner Device Info</h3>
...
+            // Map AI type_prediction to image_type select value
             if (result.type_prediction) {
-                const mapped = mapAIType(result.type_prediction);
-                const recordTypeSelect = document.getElementById('record-type');
-                if (mapped && mapped.recordType) {
-                    recordTypeSelect.value = mapped.recordType;
-                    state.aiImageTypeCode = mapped.imageType;
-                    state.aiPatientOrientation = mapped.patientOrientation;
+                let id = state.valuesets.image_types.find(x => x.display === result.type_prediction).id;
+                console.log(id);
```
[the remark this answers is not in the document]
aspiringLich: removed

bfd9000_web/archive/templates/archive/scan.html
```diff
@@ -1089,16 +1044,40 @@ <h3 class="font-bold text-lg mb-2">Scanner Device Info</h3>
...
-            }
-            if (state.aiPatientOrientation && state.aiPatientOrientation.length === 2) {
-                formData.append('patient_orientation', state.aiPatientOrientation[0]);
-                formData.append('patient_orientation', state.aiPatientOrientation[1]);
+            // FIXME: currently this is sorta hardcoded because we only support dental scans rn but this should
+            //        be updated with more logic if we support more types of scans
+            // Find any modality where x.id == 'RG'
+
+            const set_record_type_and_modality = (record_type, modality) => {
+                console.assert(state.valuesets.record_types.find(x => x.id === record_type), `Expected record type ${record_type} to exist is valueset record_types`);
```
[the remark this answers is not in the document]
aspiringLich: fixed

bfd9000_web/archive/templates/archive/scan.html
```diff
@@ -698,14 +659,13 @@ <h3 class="font-bold text-lg mb-2">Scanner Device Info</h3>
         options.forEach(opt => {
             const option = document.createElement('option');
             option.value = opt.id;
-            option.textContent = opt.display;
+            option.textContent = `${opt.display} (${opt.id})`;
```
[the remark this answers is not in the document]
aspiringLich: I don't wanna

bfd9000_web/archive/templates/archive/scan.html
```diff
@@ -138,22 +138,12 @@ <h2 class="card-title text-lg">Record Metadata</h2>
                             min-width: 8rem;
                         }
                     </style>
-                    <!-- Record Type -->
+                    <!-- Image Type -->
                     <div class="form-control">
                         <label class="label">
-                            <span class="label-text">Record Type <span class="text-error">*</span></span>
+                            <span class="label-text">Image Type<span class="text-error">*</span></span>
```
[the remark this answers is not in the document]
aspiringLich: fixed

bfd9000_web/archive/templates/archive/scan.html
```diff
@@ -1089,16 +1044,40 @@ <h3 class="font-bold text-lg mb-2">Scanner Device Info</h3>
...
+            // FIXME: currently this is sorta hardcoded because we only support dental scans rn but this should
+            //        be updated with more logic if we support more types of scans
+            // Find any modality where x.id == 'RG'
+
+            const set_record_type_and_modality = (record_type, modality) => {
+                console.assert(state.valuesets.record_types.find(x => x.id === record_type), `Expected record type ${record_type} to exist is valueset record_types`);
+                formData.append('record_type', record_type);
+                console.assert(state.valuesets.modalities.find(x => x.id === modality), `Expected modality ${modality} to exist in valueset modalities`);
+                formData.append('modality', modality);
+                console.log(`Record type: ${record_type}, Modality: ${modality}`);
```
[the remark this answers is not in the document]
aspiringLich: fixed

bfd9000_web/archive/templates/archive/scan.html
```diff
@@ -1046,17 +1001,17 @@ <h3 class="font-bold text-lg mb-2">Scanner Device Info</h3>
...
-                if (mapped && mapped.recordType) {
-                    recordTypeSelect.value = mapped.recordType;
-                    state.aiImageTypeCode = mapped.imageType;
-                    state.aiPatientOrientation = mapped.patientOrientation;
+                let id = state.valuesets.image_types.find(x => x.display === result.type_prediction).id;
+                console.log(id);
+
+                const imageTypeSelect = document.getElementById('image-type');
+                if (id) {
+                    imageTypeSelect.value = id;
```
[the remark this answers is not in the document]
aspiringLich: addressed in #72

bfd9000_web/archive/templates/archive/scan.html
```diff
@@ -1089,16 +1044,40 @@ <h3 class="font-bold text-lg mb-2">Scanner Device Info</h3>
...
+                case 'H': set_record_type_and_modality('39714003', 'RG'); break;
+                // Cephalogram, Radiographic imaging
+                case 'L': set_record_type_and_modality('201456002', 'RG'); break;
+                // Pelvis X-ray, Radiographic imaging
+                case 'P': set_record_type_and_modality('268425006', 'RG'); break;
+                // Dental model, 3D Manufacturing Modeling System
+                case 'SM': set_record_type_and_modality('302189007', 'M3D'); break;
+                case 'FM':
+                case 'F':
+                    throw new Error('Image type has no match in record type codings');
```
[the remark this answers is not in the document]
aspiringLich: I think this is ok but feel free to disagree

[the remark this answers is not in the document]
zgypa: Complete mapping added.

zgypa (commented):
**[claude-sonnet-4.6 / @worker]**

Overall the PR is clean and the direction is correct. A few issues worth addressing before merge:

**1. Network call in `initialize` blocks on external FHIR endpoint (medium)**

`import_valuesets --all` now runs unconditionally inside `initialize` when `--skip-migrate` is not set. If the FHIR endpoint at `http://terminology.open-ortho.org/...` is unreachable (offline dev, CI without network), the entire init fails with an unhandled `URLError`. There is no `--skip-valueset-import` flag and no fallback. Consider wrapping in try/except with a warning, or adding a skip flag.

**2. `_fetch_valueset` decodes as ASCII, silently drops non-ASCII (low)**

`response.read().decode("ascii", errors="ignore")` silently strips non-ASCII bytes (e.g. accented display names). FHIR JSON is UTF-8; use `"utf-8"` or read the `Content-Type` charset.

**3. AI confidence display crashes if `type_probability` is missing (medium)**

`scan.html` line 1034 reads `result.type_probability.toFixed(2)`. After this PR, `queryAI` returns `{ ...result, class_code, class_prediction, class_probability }` — if `type_probability` is absent from the AI response this throws `TypeError` in the `catch` block, showing an error to the user even though classification succeeded. Should use `result.class_probability ?? result.type_probability` (or `?.toFixed`).

**4. AI code lookup cascade has wrong ordering (low)**

Line 1013 of `scan.html`:
`result.code || result.record_type || result.prediction || result.class_code || result.class_prediction`

`result.prediction` is the raw text label (e.g. `"Lateral"`), not a valueset code, so it silently fails the `find` check — but only when `result.code` and `result.record_type` are both falsy. `class_code` (newly added this PR, the intended code) is placed *after* `result.prediction`, meaning a non-null `prediction` string shadows it. Code fields should come before text-label fields.

**5. Behavioral change: patient orientation now set by `rt_coding.code == 'L'` not `image_type.code == 'L'` (worth confirming)**

`serializers.py:578` changed from checking `image_type.code` to `rt_coding.code`. This should be equivalent per the new 1-to-1 mapping, but it is worth explicitly confirming that no record type with code `L` can represent anything other than a lateral cephalogram, since `['A', 'F']` will now be auto-applied unconditionally for any record type `L`.

bfd9000_web/archive/management/commands/initialize.py
```diff
@@ -75,6 +77,13 @@ def handle(self, *args, **options) -> None:
             self.stdout.write(self.style.NOTICE("Running migrate..."))
             call_command("migrate", verbosity=verbosity)
 
+            self.stdout.write(self.style.NOTICE("Importing all valuesets..."))
+            call_command(
+                "import_valuesets",
+                "--all",
+                verbosity=verbosity,
+            )
```
zgypa: **[claude-sonnet-4.6 / @worker]** The `initialize` command now runs `import_valuesets --all` unconditionally (whenever `--skip-migrate` is not set). This makes a network request to `http://terminology.open-ortho.org/...`. If that endpoint is unreachable (offline dev environment, CI runner without outbound network, air-gapped deploy), the init command will fail with an unhandled `URLError` before the superuser or data import steps run. There is no `--skip-valueset-import` flag and no fallback. Consider wrapping in a try/except that emits a warning and continues, or adding a dedicated skip flag.

bfd9000_web/archive/management/importers/valuesets.py
```diff
@@ -0,0 +1,110 @@
...
+    """
+    payload = _fetch_valueset(expand_url)
+    valueset = _upsert_valueset(payload, slug)
+    codings = _upsert_codings(valueset, payload)
+    _sync_valueset_links(valueset, codings)
+    return len(codings)
+
+def _fetch_valueset(url: str) -> Dict[str, Any]:
+    with urlopen(url) as response:
+        raw = response.read().decode("ascii", errors="ignore")
```
zgypa: **[claude-sonnet-4.6 / @worker]** `errors="ignore"` silently drops any non-ASCII bytes. FHIR JSON is specified as UTF-8, and display names may include accented characters. Use `decode("utf-8")` (or honour the `Content-Type` charset) to avoid silently corrupting display strings.

bfd9000_web/archive/templates/archive/scan.html
```diff
@@ -718,9 +681,8 @@ <h3 class="font-bold text-lg mb-2">Scanner Device Info</h3>
         const hasEncounter = state.encounter !== null;
         const hasFile = state.file !== null;
         const hasRecordType = document.getElementById('record-type').value !== '';
-        const hasModality = document.getElementById('modality').value !== '';
 
-        const canSubmit = hasSubject && hasEncounter && hasFile && hasRecordType && hasModality;
+        const canSubmit = hasSubject && hasEncounter && hasFile && hasRecordType;
         document.getElementById('submit-button').disabled = !canSubmit;
```
zgypa, line 692: **[claude-sonnet-4.6 / @worker]** `result.type_probability` may be `undefined` here. After this PR `queryAI` spreads the original `result` but stores the probability as `class_probability`. If the AI service does not return a `type_probability` field, calling `.toFixed(2)` throws `TypeError`, which is caught by the outer catch and shown as an error to the user even though classification succeeded. Fix: `(result.class_probability ?? result.type_probability)?.toFixed(2) ?? 'N/A'`.

bfd9000_web/archive/templates/archive/scan.html
```diff
@@ -965,8 +925,6 @@ <h3 class="font-bold text-lg mb-2">Scanner Device Info</h3>
 
         state.file = file;
         state.transformOps = [];
```
zgypa, line 932: **[claude-sonnet-4.6 / @worker]** The cascade `result.code || result.record_type || result.prediction || result.class_code || result.class_prediction` has a priority problem. `result.prediction` is the raw text label from the first AI call (e.g. `"Lateral"`), not a valueset code. It will always silently fail the `find(x => x.id === aiCode)` check, but it shadows `result.class_code` — the field added in this PR that actually contains the code. If `result.prediction` is a non-empty string and `result.code`/`result.record_type` are both absent, `class_code` is never reached. Reorder: put code fields (`class_code`, `class_prediction`) before text-label fields, or remove `result.prediction` from this branch entirely since it is handled by the `else if (result.type_prediction)` fallback below.

bfd9000_web/archive/serializers.py
```diff
@@ -564,7 +575,7 @@ def create(self, validated_data: Dict[str, Any]) -> Record:
         if request and getattr(request, 'user', None) and request.user.is_authenticated:
             scan_operator = request.user
 
-        if patient_orientation is None and image_type and getattr(image_type, 'code', None) == LATERAL_IMAGE_TYPE_CODE:
+        if patient_orientation is None and getattr(rt_coding, 'code', None) == LATERAL_RECORD_TYPE_CODE:
```
zgypa, line 578: **[claude-sonnet-4.6 / @worker]** This changes the patient-orientation auto-assignment from checking `image_type.code == 'L'` to `rt_coding.code == 'L'`. The two systems use the same letter `L` but under different code system URLs (`SYSTEM_IDENTIFIER_IMAGE_TYPE` vs `SYSTEM_RECORD_TYPE`). Per `RECORD_TYPE_MODALITY_MAP` the mapping is 1-to-1, so this should be equivalent — but it is worth confirming explicitly that `L` in the new CWRU record-type valueset exclusively means lateral cephalogram and cannot be reused for another type, since `['A', 'F']` will now be applied to every record with `record_type.code == 'L'` regardless of client input.

bfd9000_web/archive/management/commands/initialize.py
zgypa: Wrap and emit warning. no need for extra flag.

bfd9000_web/archive/management/importers/valuesets.py
zgypa: Honour content-type or FHIR JSON specification. If FHIR JSON only support UTF-8, then you can hard code UTF-8. Chose most robust solution.

bfd9000_web/archive/templates/archive/scan.html
zgypa: OK, apply.

bfd9000_web/archive/templates/archive/scan.html
zgypa: Raw result is now a 1/2 character string. Pydantic should be enforcing that in the API of it engine. Choose most simple and robust solution. Assume AI endpoint is returning valid data. The user has to be notified if the classification failed, so they can classify manually.

bfd9000_web/archive/serializers.py
zgypa: The patient orientation logic should be defined like this: The AI offers a recommendaton for flipping and/or rotating: the user confirms it, or corrects it; we store the net sum of the AI + the user flip/rotation; for each image type, we know what the orientation is that the AI is aiming for, so we should be able to calculate it; so we can show it next to the thumbnail on the right side, and on the bottom, as a dropdown, that the use can correct; now we finally can store the rotation and flip action along with the orientation vectors in the Records field. The meaning will be: "The current thumbnail is properly rotated and PatientOrientation applies. For the source image, the PatientOrientation must be reverse calculated applying the flip/rotation vectors.".

The logic for this rotation is slightly complex, and needs to go in the proper place (a service? utils?), s.t. it can be used to "encode" and "decode" PatientOrientation. Maybe the records model could have a `get_patient_orientation(thumbnail|source) -> PatientOrientation` which can be used by the frontend to populate the four letters around the image, and can be used to send to the DICOM conversion.

All this should go in a new issue. Add issue for this, make it a Task.

At the moment, we can't deal with this, so we can just add two dropdowns next to the image preview in scan, with values "Anterior,Posterior,Head,Feet,Left,Right" on each, then hard code L -> A,F and F -> L,F.

bfd9000_web/archive/management/commands/initialize.py
zgypa: **[github-copilot/claude-sonnet-4.6]** Fixed: wrapped the `import_valuesets --all` call in a try/except block. On failure a Django warning is emitted and initialization continues. No extra CLI flag added.

bfd9000_web/archive/management/importers/valuesets.py
zgypa: **[github-copilot/claude-sonnet-4.6]** Fixed: changed `decode("ascii", errors="ignore")` to `decode("utf-8")`. FHIR JSON is always UTF-8 per spec, so hardcoding UTF-8 is both correct and the most robust solution.

bfd9000_web/archive/templates/archive/scan.html
zgypa: **[github-copilot/claude-sonnet-4.6]** Applied: probability display now uses `(result.class_probability ?? result.type_probability)?.toFixed(2) ?? 'N/A'` to safely handle the case where `type_probability` is absent.

bfd9000_web/archive/templates/archive/scan.html
zgypa: **[github-copilot/claude-sonnet-4.6]** Simplified the cascade to `result.class_code || result.code`. Since the AI endpoint (Pydantic-validated) returns valid data, this is the most direct solution. If both fields are absent, a visible UI warning is shown instructing the user to classify manually.

bfd9000_web/archive/serializers.py
zgypa: **[github-copilot/claude-sonnet-4.6]** Acknowledged. This describes a fuller orientation workflow (AI recommendation → user confirm/correct → store net flip/rotation, display next to thumbnail as a correctable dropdown). This is out of scope for PR #70. Tracking as a separate follow-up task.

zgypa: **[github-copilot/claude-sonnet-4.6]** Created issue #78 to track the full patient orientation encoding/decoding service: https://github.com/open-ortho/edu.case.BFD9000/issues/78

PR #84, opened 2026-03-09
Feature/39 box integration

aspiringLich:
Upload files from MEDIA_ROOT/uploads to box. 

Note that I provide options to use jwt and developer token authentication. If the envar `BOX_DEVELOPER_TOKEN` exists it uses that, otherwise it tries to authenticate with jwt.


aspiringLich:
oh wait this should be for develop

PR #85, opened 2026-03-09
Feature/39 box integration

### LINKED ISSUE #39: Box Integration
We are planning on using Box in some respect for storage, either to store DICOM images or as a backup.

Currently we do not have a server with enough storage space to store all the images locally. Also, all the current scans are stored in box. For now, it may be necessary to use box to store all the scans. In the future, if we have access to a server with enough storage space, box could used as a backup server.

### LINKED ISSUE #43: Develop Initial Storage Layer
- With ABC interface, and ability for multiple implementations (DICOM node, box, drive, local storage/smb...). 
- This is where the original files go, once they are archived
- Once archived, they should be removed
- folder structure logic should be addressed in the implementation of this ABC, if necessary.
- First implementation should have capabilities for box. So #39 should be using this.

## Abilities

- upload new file: `upload(record:Record,destination:Endpoint)`
- list all files, with filters (search): `list(endpoint:Endpoint, filter) -> List[Record]`
- download file: `download(record:Record, from:Endpoint) -> Record` this should download the file back into local Django model.
- NO delete/rename: these operations should be done manually on the storage itself, if necessary

aspiringLich:
Upload files from MEDIA_ROOT/uploads to box in a background task. Adds Storage backends with the following methods (the implementation differs slightly from #43):

- `upload(file, path: str) -> uri: str`
- `list(path: str) -> uris: list[str]`
- `download(uri: str) -> (file, filename: str)`

The Box backend uses caching to avoid repeated redundant api calls.

Note that I provide options to use OAuth, jwt, and developer token authentication. If the envar `BOX_DEVELOPER_TOKEN` exists it uses that, otherwise it tries to authenticate with jwt or OAuth. 

Also adds logging support to console with pretty print for nice colors.

closes #39 
closes #43


bfd9000_web/BFD9000/settings.py
```diff
@@ -37,6 +49,9 @@
 if not DEBUG and 'SECRET_KEY' not in os.environ:
     raise RuntimeError('SECRET_KEY must be set when DEBUG=False')
 
+if not DEBUG and 'BOX_ACCESS_TOKEN' not in os.environ:
+    raise RuntimeError('BOX_ACCESS_TOKEN must be set when DEBUG=False')
+
```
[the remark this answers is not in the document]
aspiringLich: done

bfd9000_web/archive/media_upload.py
```diff
@@ -0,0 +1,275 @@
...
+    return files_processed
+
+
+def handle_media_file(file_path: Path) -> bool:
+    """Handle a single media file: upload, delete, and prune directory."""
+    try:
+        logger.debug(f"Handling media file: {file_path}")
+        # Attempt to upload the file
+        if upload_file(file_path):
+            # Delete the file after successful upload TODO: delete it
```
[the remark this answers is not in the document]
aspiringLich: done

bfd9000_web/requirements.txt
```diff
@@ -9,5 +9,9 @@ drf-spectacular>=0.27,<1.0
 gunicorn>=21.0,<22.0
 whitenoise>=6.6,<7.0
 openpyxl>=3.1,<4.0
+django-stubs>=4.2,<5.0
+djangorestframework-stubs>=3.14,<4.0
```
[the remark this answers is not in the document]
aspiringLich: i removed them but this one is kinda weird. Do we want a `requirements-dev.txt`?

flake.nix
```diff
@@ -20,6 +20,33 @@
         let
           pkgs = import nixpkgs { inherit system; };
           python = pkgs.python311;
+
+          # Custom build for boxsdk
+          boxsdk = python.pkgs.buildPythonPackage rec {
+            pname = "boxsdk";
+            version = "10.4.0";
+            format = "setuptools";
```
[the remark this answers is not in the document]
aspiringLich: switched to box-sdk-gen

bfd9000_web/archive/media_upload.py
```diff
@@ -0,0 +1,275 @@
...
+    UploadFileAttributesParentField,
+)
+from django.conf import settings
+from dataclasses import dataclass
+
+from BFD9000.settings import (
+    BOX_ACCESS_TOKEN,
+    BOX_DEVELOPER_TOKEN,
+    BOX_FOLDER_ID,
+    BOX_JWT_CONFIG_FILE,
```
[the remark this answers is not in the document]
aspiringLich: resolved

bfd9000_web/archive/media_upload.py
```diff
@@ -0,0 +1,275 @@
...
+_item_cache: dict[tuple[str, str], ItemData | None] = {}
+
+
+def _get_box_client() -> BoxClient:
+    """Get an authenticated Box client."""
+    if BOX_JWT_CONFIG_FILE:
+        jwt_config = JWTConfig.from_config_file(config_file_path="boxjwt.json")
+        auth = BoxJWTAuth(config=jwt_config)  # pyright: ignore[reportArgumentType]
+    elif BOX_DEVELOPER_TOKEN:
+        auth = BoxDeveloperTokenAuth(token=BOX_DEVELOPER_TOKEN)
```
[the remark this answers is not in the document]
aspiringLich: done

bfd9000_web/archive/apps.py
```diff
@@ -1,9 +1,34 @@
...
 class ArchiveConfig(AppConfig):
     """Configure default settings for the archive app."""
     default_auto_field = "django.db.models.BigAutoField"  # pyright: ignore[reportAssignmentType]
     name = "archive"
+
+    def ready(self):
+        """Initialize the archive app and start background tasks."""
+        # Only start during runserver, not during migrations, tests, etc.
+        # Guard against running twice in development (autoreloader issue)
+        if 'runserver' in sys.argv and not self._is_reloader_process():
```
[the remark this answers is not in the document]
aspiringLich: done with some hardcoding goodness

bfd9000_web/archive/media_upload.py
```diff
@@ -0,0 +1,275 @@
...
+                        del _item_cache[cache_key]
+                        logger.debug(f"Invalidated cache for deleted file: {file_name}")
+                else:
+                    logger.error(f"Preflight check failed: {e}")
+                    return False
+
+        # Upload the file
+        with open(file_path, 'rb') as file_stream:
+            if file_size > 50_000_000:
+                session_id = upload_url.upload_url.rsplit('?', 1)[1]  # pyright: ignore[reportOptionalMemberAccess]
```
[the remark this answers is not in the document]
aspiringLich: multiple 409's should be impossible! ill add a check though

bfd9000_web/BFD9000/settings.py
```diff
@@ -37,6 +49,9 @@
 if not DEBUG and 'SECRET_KEY' not in os.environ:
     raise RuntimeError('SECRET_KEY must be set when DEBUG=False')
 
+if not DEBUG and 'BOX_ACCESS_TOKEN' not in os.environ:
```
zgypa: [claude-sonnet-4.6] **[BUG - HIGH]** Wrong environment variable in production startup check.

`BOX_ACCESS_TOKEN` is never defined or used anywhere in the codebase. The actual variables are `BOX_DEVELOPER_TOKEN` and `BOX_JWT_CONFIG_FILE`. This check will always pass in production (since `BOX_ACCESS_TOKEN` is never set), silently allowing the app to start without Box credentials. The worker then fails at runtime instead of at startup.

Fix: change `BOX_ACCESS_TOKEN` to check `BOX_DEVELOPER_TOKEN` and/or `BOX_JWT_CONFIG_FILE`.

bfd9000_web/archive/media_upload.py
```diff
@@ -0,0 +1,282 @@
...
+
+def _get_box_client() -> BoxClient:
+    """Get an authenticated Box client."""
+    if BOX_JWT_CONFIG_FILE:
+        jwt_config = JWTConfig.from_config_file(config_file_path=BOX_JWT_CONFIG_FILE)
+        auth = BoxJWTAuth(config=jwt_config)  # pyright: ignore[reportArgumentType]
+    elif BOX_DEVELOPER_TOKEN:
+        auth = BoxDeveloperTokenAuth(token=BOX_DEVELOPER_TOKEN)
+
+    if auth is None:
```
zgypa: [claude-sonnet-4.6] **[BUG - HIGH]** `NameError` instead of `RuntimeError` in `_get_box_client`.

If neither `BOX_JWT_CONFIG_FILE` nor `BOX_DEVELOPER_TOKEN` is set, `auth` is never assigned before `if auth is None` is evaluated, causing `NameError: name 'auth' is not defined` instead of the intended `RuntimeError`. The `media_upload_worker` guard prevents this in normal use, but any direct call to `process_media_files()` or `upload_file()` (e.g., from a management command or test) will get an unhelpful `NameError`.

Fix: initialize `auth = None` at the top of the function before the if/elif branches.

bfd9000_web/archive/media_upload.py
```diff
@@ -0,0 +1,282 @@
...
+
+                    # Invalidate cache for this file
+                    cache_key = (current_folder_id, file_name)
+                    if cache_key in _item_cache:
+                        del _item_cache[cache_key]
+                        logger.debug(f"Invalidated cache for deleted file: {file_name}")
+                else:
+                    logger.error(f"Preflight check failed: {e}")
+                    return False
+        if upload_url is None:
```
zgypa: [claude-sonnet-4.6] **[BUG - MEDIUM]** Retry loop can exit with `upload_url = None` after 3 attempts.

The comment says "multiple 409 responses should be impossible," but this is a race condition. If another process uploads a file with the same name between the delete and the next preflight check, two consecutive 409s can exhaust all 3 retries. After exhausting retries, `RuntimeError` is raised, caught by the outer `except Exception`, logged, and the function returns `False`. The file will not be retried until the next worker cycle (60s later). The assertion in the error message is incorrect.

bfd9000_web/archive/apps.py
```diff
@@ -1,9 +1,39 @@
...
     name = "archive"
+
+    def ready(self):
+        """Initialize the archive app and start background tasks."""
+        # Guard against running twice in development (autoreloader issue)
+        if 'runserver' in sys.argv:
+            if not self._is_reloader_process():
+                self._start_background_task()
+        # are we in production (gunicorn)?
+        elif sys.argv[0] == 'gunicorn':
```
zgypa: [claude-sonnet-4.6] **[ISSUE - MEDIUM]** Gunicorn detection is fragile.

When gunicorn is invoked via its full path (e.g. `/home/user/.venv/bin/gunicorn`) or as a module (`python -m gunicorn`), `sys.argv[0]` will not equal `'gunicorn'`. The background task will silently not start in production, with only an INFO log message.

Fix: use `os.path.basename(sys.argv[0]) == 'gunicorn'` or check for a `SERVER_SOFTWARE` env var set by gunicorn.

bfd9000_web/requirements.txt
```diff
@@ -9,5 +9,9 @@ drf-spectacular>=0.27,<1.0
 gunicorn>=21.0,<22.0
 whitenoise>=6.6,<7.0
 openpyxl>=3.1,<4.0
+django-stubs>=4.2,<5.0
```
zgypa: [claude-sonnet-4.6] **[ISSUE - MEDIUM]** Type stubs belong in `requirements-dev.txt`, not `requirements.txt`.

Per `AGENTS.md`: "django-types should be available in the development environment; it is a dev-time dependency, not a production runtime requirement." `django-stubs` and `djangorestframework-stubs` are pure type-checking tools and add unnecessary packages to the production environment.

Fix: move both to `requirements-dev.txt`.

bfd9000_web/archive/media_upload.py
```diff
@@ -0,0 +1,282 @@
...
+            raise RuntimeError("Internal Error: expected upload_url to be set, multiple 409 responses should be impossible")
+
+        # Upload the file
+        with open(file_path, 'rb') as file_stream:
```
zgypa: [claude-sonnet-4.6] **[MINOR]** File is opened before the size check that raises `NotImplementedError`.

The file is opened with `open(file_path, "rb")` before checking `if file_size > 50_000_000`, which immediately raises `NotImplementedError`. The file handle is opened and closed for no reason. Additionally, `session_id` on the next line is dead code — it is assigned and immediately discarded by the raise.

Fix: move the size check before the `with open(...)` block.

bfd9000_web/requirements.txt
```diff
@@ -9,5 +9,9 @@ drf-spectacular>=0.27,<1.0
...
+djangorestframework-stubs>=3.14,<4.0
+python-dotenv>=1.2.1
```
zgypa: [claude-sonnet-4.6] **[MINOR]** `python-dotenv` has no upper version bound.

All other packages in `requirements.txt` specify upper bounds (e.g. `>=1.0,<2.0`). Missing an upper bound on `python-dotenv` means a future major release could silently break the build.

Consider: `python-dotenv>=1.2.1,<2.0`

bfd9000_web/BFD9000/settings.py
aspiringLich: this check doesnt even need to happen. the background task is optional anyway

bfd9000_web/archive/media_upload.py
aspiringLich: done

bfd9000_web/archive/media_upload.py
aspiringLich: changed the assertion

bfd9000_web/archive/media_upload.py
aspiringLich: its configured this way because a future implementation will have this structure

bfd9000_web/requirements.txt
aspiringLich: done

bfd9000_web/requirements.txt
aspiringLich: done

bfd9000_web/archive/apps.py
aspiringLich: fair enough

bfd9000_web/requirements.txt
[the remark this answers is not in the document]
zgypa: @aspiringLich Of course, whiy not ? requiremetns-dev are tools required for developmetn, like linting tools and stubs. By the way, we use Pylance, which does not usee *stubs, it uses the other ones, so we shouldn't need *stubs here. Look into this.

zgypa:
[claude-sonnet-4-6] **[REQUIRED] Issue #43 ("Develop Initial Storage Layer") is not fulfilled by this implementation**

This PR is tagged as closing both #39 and #43. The Box upload worker satisfies the spirit of #39, but #43 specifies a different design:

> _With ABC interface, and ability for multiple implementations (DICOM node, box, drive, local storage/smb…). First implementation should have capabilities for box. So #39 should be using this._

Required abilities per #43:
- `upload(record: Record, destination: Endpoint)`
- `list(endpoint: Endpoint, filter) -> List[Record]`
- `download(record: Record, from: Endpoint) -> Record`

What this PR delivers instead:
- A **hardcoded Box background worker** with no ABC and no pathway to other implementations without rewriting the module.
- `upload` is implicit (background file scanner), not a callable interface method.
- `list` is **not implemented at all**.
- `download_file(box_file_id: str) -> (IO[bytes], str)` does not match the required signature and returns a stream, not a `Record`.

**Two acceptable paths forward:**
1. Refactor `media_upload.py` around an ABC storage interface before merge, so that #43 is genuinely closed.
2. Explicitly scope this PR to only close #39, remove #43 from the closing list, and open a follow-up issue/PR for the ABC layer.

bfd9000_web/requirements.txt
```diff
@@ -9,6 +9,8 @@ drf-spectacular>=0.27,<1.0
 gunicorn>=21.0,<22.0
 whitenoise>=6.6,<7.0
 openpyxl>=3.1,<4.0
+python-dotenv>=1.2.1,<2.0
+boxsdk>=10.4.0
```
zgypa: [claude-sonnet-4-6] **[BLOCKER] Wrong Box SDK package**

The code imports from `box_sdk_gen` (Box Gen2 Python SDK):

```python
from box_sdk_gen import BoxAPIError, BoxClient, BoxJWTAuth, ...
```

But this line installs `boxsdk` — the old, completely incompatible SDK. A clean `pip install -r requirements.txt` will produce `ModuleNotFoundError: No module named 'box_sdk_gen'` at startup.

The author replied "switched to box-sdk-gen" to the Copilot comment but the change was never applied in the diff.

**Fix:** replace with `box-sdk-gen>=1.0,<2.0` (check PyPI for the current stable version and add an upper bound consistent with other entries here). The same issue exists in `flake.nix` lines 24–59 — see the comment there.

flake.nix
```diff
@@ -28,6 +55,8 @@
             buildInputs = [
               python
               python.pkgs.pip
+              python.pkgs.python-dotenv
+              boxsdk
```
zgypa: [claude-sonnet-4-6] **[BLOCKER] `flake.nix` builds `boxsdk`, not `box_sdk_gen`**

Same issue as `requirements.txt`: this block builds and installs the old `boxsdk` package, but the code imports from `box_sdk_gen`. Local Nix development will fail to start with a `ModuleNotFoundError`.

The author replied "switched to box-sdk-gen" to the Copilot comment but this block was not updated.

**Fix:** replace with a derivation for `box-sdk-gen`, or drop the custom build and let it be installed via pip from `requirements.txt` like the other packages.

bfd9000_web/archive/apps.py
```diff
@@ -1,9 +1,39 @@
...
+                self._start_background_task()
+        # are we in production (gunicorn)?
+        elif os.path.basename(sys.argv[0]) == 'gunicorn':
+            self._start_background_task()
+        else:
+            logger.info("Background tasks not started: conditions have not been met (runserver, gunicorn)")
+
+    def _is_reloader_process(self):
+        """Check if this is the reloader process (development only)."""
+        return os.environ.get('RUN_MAIN') != 'true'
```
zgypa: [claude-sonnet-4-6] **[BLOCKER] `_is_reloader_process` logic is confusing and the name is misleading**

```python
return os.environ.get('RUN_MAIN') != 'true'
```

Django's autoreloader sets `RUN_MAIN=true` in the **child** (real server) process. The **parent** (reloader) does NOT set it. So this method returns `True` for the reloader and `False` for the real server — the opposite of what the name implies. The call site `if not self._is_reloader_process()` introduces a second negation that makes the control flow hard to follow.

The net behavior is coincidentally correct (task starts in the child), but any future reader will need to work through the double-negation to verify that, making this a maintenance hazard.

**Fix:** rename to `_is_main_server_process` and flip the return value, then drop the `not` at the call site:

```python
def _is_main_server_process(self):
    return os.environ.get('RUN_MAIN') == 'true'

# in ready():
if 'runserver' in sys.argv:
    if self._is_main_server_process():
        self._start_background_task()
```

bfd9000_web/archive/media_upload.py
```diff
@@ -0,0 +1,307 @@
...
+class ItemData:
+    id: str
+    type: FileBaseTypeField | FolderBaseTypeField | WebLinkBaseTypeField
+
+_item_cache: dict[tuple[str, str], ItemData | None] = {}
+
+
+def _get_box_client() -> BoxClient:
+    """Get an authenticated Box client."""
+    if BOX_JWT_CONFIG_FILE:
```
zgypa: [claude-sonnet-4-6] **[MEDIUM] Auth order contradicts PR description**

The PR description states: _"If the envar `BOX_DEVELOPER_TOKEN` exists it uses that, otherwise it tries to authenticate with JWT."_

The code does the opposite — JWT wins when both are set. For local development the developer token is the simpler path and should take priority.

**Fix:** swap the branches:

```python
if BOX_DEVELOPER_TOKEN:
    auth = BoxDeveloperTokenAuth(token=BOX_DEVELOPER_TOKEN)
elif BOX_JWT_CONFIG_FILE:
    jwt_config = JWTConfig.from_config_file(config_file_path=BOX_JWT_CONFIG_FILE)
    auth = BoxJWTAuth(config=jwt_config)  # pyright: ignore[reportArgumentType]
else:
    raise RuntimeError(...)
```

bfd9000_web/archive/media_upload.py
```diff
@@ -0,0 +1,307 @@
+"""Background worker for uploading media files."""
+
+import logging
+import time
+from pathlib import Path
+from typing import IO, List
+
+from box_sdk_gen import BoxAPIError, BoxClient, BoxJWTAuth, CreateFolderParent, FileBaseTypeField, FolderBaseTypeField, JWTConfig, WebLinkBaseTypeField
```
zgypa: [claude-sonnet-4-6] **[MEDIUM] No tests for the new upload/retry/cleanup logic**

`archive/tests/` already exists with a pytest setup, but this entire new module has no test coverage. Copilot flagged this with no response from the author.

Minimum cases needed:
- Happy-path upload → `source_file` updated to `box://<id>` in the DB
- 409-conflict → delete-and-retry path succeeds on second attempt
- Exhausted retries (3× 409) → returns `False`, local file not deleted
- `prune_empty_directory` → removes empty parents, stops at `MEDIA_ROOT`
- `_get_item_by_name` → positive and negative cache hits avoid extra API calls
- `media_upload_worker` → exits cleanly when env vars are missing

bfd9000_web/archive/media_upload.py
```diff
@@ -0,0 +1,307 @@
...
+                        del _item_cache[cache_key]
+                        logger.debug(f"Invalidated cache for deleted file: {file_name}")
+                else:
+                    logger.error(f"Preflight check failed: {e}")
+                    return None
+        if upload_url is None:
+            raise RuntimeError("Internal Error: multiple 409 responses recieved when trying to upload file.")
+
+        # Upload the file
+        with open(file_path, 'rb') as file_stream:
```
zgypa: [claude-sonnet-4-6] **[MINOR] File opened before the size check that immediately raises**

The file handle is acquired on this line, then the very next line raises `NotImplementedError` for files > 50 MB. The open is wasted, and `session_id` on the following line is dead code (immediately discarded by the raise). @zgypa flagged this and the author replied "done" — but the fix was not applied in the diff.

**Fix:** hoist the guard above the `with` block:

```python
if file_size > 50_000_000:
    raise NotImplementedError("TODO: implement chunked uploads for files > 50MB")

with open(file_path, 'rb') as file_stream:
    result = client.uploads.upload_file(...)
```

bfd9000_web/archive/views.py
```diff
@@ -343,8 +344,14 @@ def image(self, request: Request, pk: Optional[int] = None, **kwargs: Any) -> An
         digital_record = self.get_object()
         if not getattr(digital_record, 'source_file', None):
             return Response({"error": "No image file available"}, status=404)
-        source_file = digital_record.source_file
-        return FileResponse(source_file.open('rb'))
+        source_file_name = digital_record.source_file.name
+        if source_file_name.startswith("box://"):
+            import mimetypes
```
zgypa: [claude-sonnet-4-6] **[MINOR] `import mimetypes` inside a function body**

Standard Python convention (and most linters) require imports at the top of the module. Move `import mimetypes` to the top-level import block in `views.py`.

bfd9000_web/BFD9000/settings.py
```diff
@@ -242,8 +254,63 @@ def _prefix_url(path):
 
 SCANNER_API_BASE = os.environ.get('SCANNER_API_BASE', 'http://localhost:5000')
 SCANNER_DEVICE_ID = os.environ.get('SCANNER_DEVICE_ID', 'scanner-001')
-BFD9020_BASE_URL = os.environ.get(
-    'BFD9020_BASE_URL', 'https://wingate.case.edu/bfd9020')
+BFD9020_BASE_URL = os.environ.get('BFD9020_BASE_URL', 'https://wingate.case.edu/bfd9020')
+
+# Logging Configuration
+class PrettyFormatter(logging.Formatter):
```
zgypa, line 284: [claude-sonnet-4-6] **[LOW] `PrettyFormatter` hardcodes ANSI codes unconditionally**

In Docker or any log-aggregation pipeline (CloudWatch, Datadog, Loki, etc.) these control characters render literally, making logs harder to search and parse.

Suggested fix — enable colors only when the output is a TTY:

```python
def format(self, record):
    use_color = hasattr(sys.stderr, 'isatty') and sys.stderr.isatty()
    log_fmt = self.FORMATS.get(record.levelno) if use_color else         '%(asctime)s %(levelname)-5s %(name)s: %(message)s'
    return logging.Formatter(log_fmt).format(record)
```

bfd9000_web/.env.example
```diff
@@ -0,0 +1,19 @@
+# Django Settings
```
zgypa, line 1: [claude-sonnet-4-6] **[LOW] Old `dot-env.example` removed but developer docs not updated**

The old example env file (`dot-env.example`) is deleted and replaced by `.env.example`, but any onboarding instructions, wiki pages, or setup scripts that tell developers to copy `dot-env.example` will now silently produce an empty environment, causing confusing startup failures.

Please check README, wiki pages, and setup scripts for references to `dot-env.example` and update them to point to `.env.example`.

bfd9000_web/archive/media_upload.py
```diff
@@ -0,0 +1,275 @@
...
+    except Exception as e:
+        logger.error(f"Error handling file {file_path}: {e}", exc_info=True)
+        return False
+
+
+def upload_file(file_path: Path) -> bool:
+    """
+    Upload a file to Box.com, creating folder structure as needed, and deleting the file if it already exists.
+
+    The folder structure in Box will mirror the structure relative to MEDIA_ROOT.
```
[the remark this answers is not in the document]
zgypa: [claude-sonnet-4-6] This was not addressed — no tests were added in the diff and there was no response to this comment. Still required.

bfd9000_web/.env.example
```diff
@@ -0,0 +1,19 @@
...
+SECRET_KEY=your-secret-key-here
+DEBUG=True
+DJANGO_ALLOWED_HOSTS=localhost,127.0.0.1
```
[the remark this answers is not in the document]
zgypa: [claude-sonnet-4-6] No response and not addressed in the diff. README and any setup scripts that reference `dot-env.example` still need to be updated to `.env.example`.

bfd9000_web/BFD9000/settings.py
```diff
@@ -227,8 +242,63 @@ def _prefix_url(path):
...
+    """A custom formatter to add color to stdout log records."""
+
+    GRAY = "\x1b[90m"
+    YELLOW = "\x1b[33;21m"
+    RED = "\x1b[31;21m"
+    BOLD_RED = "\x1b[31;1m"
+    RESET = "\x1b[0m"
```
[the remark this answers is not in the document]
zgypa: [claude-sonnet-4-6] No response and not addressed in the diff. `PrettyFormatter` still emits ANSI codes unconditionally. This needs a TTY check before enabling colors, otherwise production logs will contain raw control characters.

flake.nix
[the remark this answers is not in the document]
zgypa: [claude-sonnet-4-6] The author replied "switched to box-sdk-gen" but **this was not applied in the diff**. `flake.nix` still builds and installs `boxsdk` (the old SDK) while the code imports from `box_sdk_gen` (the new Gen2 SDK). These are incompatible packages — local Nix development will fail with `ModuleNotFoundError: No module named 'box_sdk_gen'`. The same mismatch exists in `requirements.txt` (line 13). Both files need to be updated.

bfd9000_web/archive/media_upload.py
zgypa: [claude-sonnet-4-6] No response and not fixed in the diff. `open(file_path, "rb")` is still called before the `if file_size > 50_000_000` check that raises `NotImplementedError`, and `session_id` is still dead code. The fix is to move the size guard above the `with open(...)` block.

zgypa:
[claude-sonnet-4-6] **Review summary — please check all threads before merging**

Several comments in this PR may be hard to spot because they are anchored to outdated diff positions and hidden behind collapsed "Show outdated" banners in the Files Changed tab. This comment summarizes all open findings for visibility.

---

**Blockers**

- `requirements.txt` line 13 + `flake.nix` lines 24–59 — **Wrong Box SDK package.** The code imports `box_sdk_gen` but both files install `boxsdk` (the old, incompatible SDK). The author replied "switched to box-sdk-gen" in the `flake.nix` thread but the change was never applied in the diff. A clean install will fail with `ModuleNotFoundError`.
- `apps.py` line 32 — **`_is_reloader_process` is misleadingly named.** The double-negation at the call site is a maintenance hazard. Rename to `_is_main_server_process` and flip the return value.

**Medium**

- `media_upload.py` line 41 — Auth order contradicts PR description (JWT wins over developer token, should be the other way around).
- `media_upload.py` (whole file) — No tests added for any of the new upload/retry/cleanup logic. The Copilot thread on this received no response.

**Minor**

- `media_upload.py` line 193 — File is opened before the `NotImplementedError` size check. zgypa flagged this; no response, not fixed in diff.
- `views.py` line 349 — `import mimetypes` inside a function body; should be top-level.

**Low**

- `settings.py` (`PrettyFormatter`) — ANSI codes emitted unconditionally; breaks log aggregation in Docker/production. Copilot thread received no response.
- `.env.example` — `dot-env.example` was removed but developer docs/scripts referencing it were not updated. Copilot thread received no response.

**Architecture**

- This PR is tagged as closing both #39 and #43. Issue #43 requires an ABC storage interface with `upload`, `list`, and `download` methods and pluggable implementations. This PR delivers a hardcoded Box worker with no ABC and no `list()` at all. Either refactor to satisfy #43 before merge, or remove #43 from the closing list and track it separately.

zgypa (changes requested):
See comments.

bfd9000_web/requirements.txt
aspiringLich: yeah yeah i forgot to push

flake.nix
aspiringLich: resolved.

bfd9000_web/.env.example
aspiringLich: updated

bfd9000_web/BFD9000/settings.py
aspiringLich: resolved

bfd9000_web/archive/views.py
aspiringLich: resolved

bfd9000_web/archive/media_upload.py
aspiringLich: resolved

bfd9000_web/archive/apps.py
aspiringLich: resolved

bfd9000_web/archive/media_upload.py
aspiringLich: resolved

aspiringLich:
Resolved all reviews; updated initial description

bfd9000_web/archive/storage.py
```diff
@@ -0,0 +1,340 @@
...
+        logger.error("Box API error creating folder %s: %s", folder_name, exc, exc_info=True)
+        return None
+    except Exception as exc:
+        logger.error("Error creating folder %s: %s", folder_name, exc, exc_info=True)
+        return None
+
+
+# ── Abstract interface ────────────────────────────────────────────────────────
+
+class StorageBackend(ABC):
```
zgypa-agent, line 151: Medium: this is closer to issue #43 than the earlier version, but it still does not match the contract that issue describes even though the PR body says it closes `#43`. The issue requires a record/endpoint abstraction (`upload(record, destination)`, `list(endpoint, filter) -> List[Record]`, `download(record, from) -> Record`), while this ABC is still path/link based and returns raw streams/strings. `archive.media_upload` also still hardcodes `BoxStorageBackend()` directly instead of depending on the abstraction for uploads. Unless the PR description is narrowed back to only `#39`, the scope mismatch is still real.

bfd9000_web/archive/media_upload.py
```diff
@@ -0,0 +1,103 @@
...
+from archive.models import DigitalRecord
+from .storage import BoxStorageBackend
+
+logger = logging.getLogger(__name__)
+
+
+def media_upload_worker():
+    from BFD9000.settings import BOX_DEVELOPER_TOKEN, BOX_FOLDER_ID, BOX_JWT_CONFIG_FILE
+
+    if not BOX_DEVELOPER_TOKEN and not BOX_JWT_CONFIG_FILE:
```
zgypa-agent: High: this startup guard is stricter than the backend auth logic below. `archive.storage._get_box_client()` supports OAuth with `BOX_OAUTH_CLIENT_ID` + `BOX_OAUTH_CLIENT_SECRET`, and the new compose/example docs advertise that as a supported configuration, but the worker exits here unless a developer token or JWT config is present. In an OAuth-only deployment, downloads can authenticate but the background uploader never starts, so files remain on local disk indefinitely.

bfd9000_web/archive/media_upload.py
```diff
@@ -0,0 +1,103 @@
...
+        if count != 1:
+            logger.error(
+                "Expected 1 record for %s, found %d; skipping DB update", relative_path, count
+            )
+            return False
+
+        with open(file_path, "rb") as f:
+            link = BoxStorageBackend().upload(f, str(relative_path))
+
+        qs.update(source_file=link)
```
zgypa-agent, line 89: High: this rewrites `DigitalRecord.source_file` from a local media path to a `box://...` link, but the rest of the API still serializes that field as if it were a local `FileField`. `DigitalRecordSerializer.get_image_url()` still returns `obj.source_file.url` and `get_file_size()` still reads `obj.source_file.size`, so after archival Django advertises `/media/box%3A/123` as the image URL and size lookup falls back to `None`/`FileNotFoundError`. I verified that behavior locally with `source_file.name = "box://123"`. The new `image()` view handles Box-backed files, but list/detail responses will still expose broken metadata unless the serializer is updated too.

bfd9000_web/archive/media_upload.py
```diff
@@ -0,0 +1,103 @@
...
+
+
+def process_media_files() -> int:
+    """Upload all pending files from the local media/uploads directory to Box."""
+    media_root = Path(settings.MEDIA_ROOT).joinpath("uploads")
+
+    if not media_root.exists():
+        return 0
+
+    image_extensions = {".jpg", ".jpeg", ".png", ".gif", ".bmp", ".tiff", ".webp"}
```
zgypa-agent, line 47: Medium: this worker only archives a small image-extension allowlist, but the rest of the app still accepts other `source_file` types. In particular, `scan_tiff_preview()` accepts both `.tif` and `.tiff`, and `DigitalRecord.source_file` is still documented as supporting `PNG/TIFF/STL`. Records uploaded as `.tif` or `.stl` will never be archived or cleaned up by this job.

bfd9000_web/archive/storage.py
```diff
@@ -0,0 +1,340 @@
...
+
+    def download(self, link: str) -> tuple[IO[bytes], str]:
+        """Delegate to whichever backend owns this link."""
+        return get_backend(link).download(link)
+
+
+# ── Factory functions ─────────────────────────────────────────────────────────
+
+def get_backend(link: str) -> StorageBackend:
+    """Return the backend that owns *link* (for download by known link)."""
```
zgypa: in the comment add that the proper backend is chosed from the URI prefix "ex: box://"

bfd9000_web/archive/storage.py
zgypa: The issue isn't that precise. E.g. you can figure out the backedn type from the URI prefix. So do that. And then use URI types instead of str for typing. That will clean up the module. 

[nitpick] if you feel like it, you can refactor the module into a package, and have each ABC and implementation as a different module.

aspiringLich:
To authenticate: Make a Box OAuth 2.0 app, copy the client and secret over, and go to `/box/oauth/start`. Sign in with box, you will be redirected and then its all good. 

Changed the file saving so now it generates a uuid for each file.

tnichols217:
Superseded by #96

PR #87, opened 2026-03-09
Fix: animated skeleton for infinite scroll does not dissapear

### LINKED ISSUE #86: Animated Skeleton for infinite scroll does not dissapear
The pulsing animated loading indicator for the infinitely scrolling tables doesnt go away when theres no more scrolling to be done

aspiringLich:
It now does go away.

Also removed it getting changed to a spinny loader thing because it looks bad.


aspiringLich:
oh i keep doing this. ugh

PR #88, opened 2026-03-09
Fix: Animated Skeleton for infinite scroll does not dissapear

aspiringLich:
It now does go away.

Also removed it getting changed to a spinny loader thing because it looks bad.


zgypa (approved):
LGTM

thanx, Needed.

PR #91, opened 2026-03-11
Feature/67 richardson

### LINKED ISSUE #67: Add Richardson Collection
Amanda or someone is scanning them at the moment, so we need them there.

Index files in box: https://cwru.app.box.com/file/1759589983019

The spreadsheet is comprehensive and should let us import:

- Collection: A new Collection for Richardson
- Subjects
- Encounters
- Physical Records

Use the following text to extract as much metadata as possible for the collection. Use all info here:

> Longitudinal records on One Hundred and Sixty (160) children, on which the study sample of Fighty-Two (82) persons
> was drawn. The sample is equally divided between the two sexes; i. e . ,Forty One (41) boys and Forty-One (41) girls.
> Records include side view X-rays of the head and plaster of Paris's [sic] casts of the upper and lower teeth.
> Each of the Eighty-Two (82) children was followed for at least eleven (11) years, from six (6) years of age or younger at
> the time that record collection began through the age of sixteen (16) years. None received orthodontic treatment before or
> during the period of record collection.
> The following records were taken on the children semiannually up to fourteen (14) years of age and annually thereafter, on
> or near the child 's birthday:
> 1
> standard lateral cephalomettic radiographs with a leaded millimeter ruler in place to facilitate checking for
> enlargement;
> 2,
> 3.
> standard postero -anterior and oblique; Forty-Five (45) cephalograms ;
> alginate impressions of the teeth and alveolar ridges with casts being poured immediately in orthodontic stone;
> 4. a complete medical history.
> No Personal Identifiable Information, including but not limited to, Protecred Health Information as defined by the Health Insurance Portability
> and Acconntability Act as amendedfrom time to time, is included in any of the above.
> Any use of Dr. Elisha R. Richardson, B.S., D.D.S., M.S., Ph.D. research materials (such as, but not limited to, atlas,
> journals, publications, x-rays, slides, dental molds, etc) in any format to be referenced to Dr. Elisha R. Richardson
> B.S.,, D.D.S., M.S., Ph.D. in all present and future use.

zgypa:
Closes #67


bfd9000_web/archive/serializers.py
```diff
@@ -652,18 +685,26 @@ def create(self, validated_data: Dict[str, Any]) -> DigitalRecord:
...
             )
 
-            # PHYSICAL RECORD: Get or create by (record_type, encounter)
-            physical_record = PhysicalRecord.objects.filter(
-                record_type=rt_coding,
-                encounter=encounter
-            ).first()
-            if not physical_record:
-                physical_record = PhysicalRecord.objects.create(
+            # PHYSICAL RECORD: Use explicitly provided one, or get/create by (record_type, encounter)
```
zgypa, line 688: openai/gpt-5.4 High: This still falls back to PhysicalRecord.objects.filter(record_type=rt_coding, encounter=encounter).first() when physical_record is omitted, but this PR removes the unique (record_type, encounter) constraint. Once multiple physical artifacts of the same type can exist on one encounter, a new scan can silently bind to an arbitrary existing artifact instead of forcing an explicit selection or creating a new one.

bfd9000_web/archive/views.py
```diff
@@ -275,6 +278,61 @@ class ArchiveLocationViewSet(viewsets.ModelViewSet):
...
+        'device',
+    ).prefetch_related(
+        'encounter__subject__identifiers',
+        'locations',
+        'identifiers',
+    )
+    serializer_class = PhysicalRecordSerializer
+    filterset_fields = ['encounter', 'record_type']
+    filter_backends = [BoltonRecordSearchFilter, filters.OrderingFilter]
+    search_fields = ['^identifiers__value']
```
zgypa, line 323: openai/gpt-5.4 Medium: Search here still only matches persisted identifiers__value, but the UI now presents the computed Bolton identifier as the primary record ID. Because this PR only adds schema and computes identifier_str on the fly, existing rows remain unsearchable by the identifier users see until the data is backfilled or each record is resaved.

bfd9000_web/archive/models.py
```diff
@@ -732,6 +848,62 @@ def subject(self) -> 'Subject':
...
+        )
+
+    def save(self, *args: Any, **kwargs: Any) -> None:
+        encounter_pk: Optional[int] = self.encounter_id  # type: ignore[attr-defined]
+        record_type_pk: Optional[int] = self.record_type_id  # type: ignore[attr-defined]
+        if self.sequence_number is None and encounter_pk and record_type_pk:
+            max_seq: Optional[int] = PhysicalRecord.objects.filter(
+                encounter_id=encounter_pk,
+                record_type_id=record_type_pk,
+            ).aggregate(m=Max('sequence_number'))['m']
```
zgypa, line 878: claude-sonnet-4.6 High: Race condition in sequence_number assignment. The MAX query and subsequent INSERT are not atomic. Under concurrent saves for the same (encounter_id, record_type_id), two transactions can both read the same max sequence_number before either has committed, resulting in duplicate sequence numbers. Duplicate sequence numbers break the uniqueness of the Bolton record identifier (e.g., two records both get sequence 01 and therefore the same bolton_record_id, then _assign_bolton_record_identifier() calls get_or_create on the same value and silently attaches the same Identifier row to both records). Fix: use SELECT FOR UPDATE on the MAX query, or add a database-level unique constraint on (encounter, record_type, sequence_number) for PhysicalRecord, and the equivalent for DigitalRecord.

bfd9000_web/archive/serializers.py
zgypa: claude-sonnet-4.6: This comment is **valid**. The fallback `.first()` call (ordered by `-acquisition_datetime`) silently picks an arbitrary `PhysicalRecord` when multiple exist for the same `(record_type, encounter)`. Prior to this PR the unique constraint guaranteed at most one match, so `.first()` was safe. Now that the constraint is removed the ambiguity is real and the severity (High) is appropriate.

bfd9000_web/archive/views.py
zgypa: claude-sonnet-4.6: This comment is **partially valid but overstated**. The claim that the Bolton identifier is only computed on the fly is incorrect: `_assign_bolton_record_identifier()` is called inside `save()` on both `PhysicalRecord` and `DigitalRecord`, which persists the value to the `identifiers` M2M. New records created after this PR *will* be findable by search immediately. The real (narrower) issue is that **pre-existing rows imported before this PR** have no Bolton identifier in the `identifiers` table yet and remain unsearchable until each is re-saved or a backfill migration is run. The severity (Medium) is appropriate for that narrower scope.

bfd9000_web/archive/models.py
```diff
@@ -841,10 +1020,71 @@ def subject(self) -> Subject:
     def __str__(self) -> str:
         return f"DigitalRecord {self.pk} ({self.record_type})"
 
-    def save(self, *args, **kwargs) -> None:
+    @property
+    def bolton_record_id(self) -> str:
+        """Compute the Bolton-style record identifier for this digital record."""
+        try:
+            encounter = self.series.imaging_study.encounter
+        except Exception:
```
zgypa: claude-sonnet-4.6 Medium: Bare `except Exception` silently swallows all errors here. If `self.series`, `self.series.imaging_study`, or `self.series.imaging_study.encounter` raises anything other than an `AttributeError` (e.g. a database error, a misconfigured relation, an unexpected `None` mid-chain), the failure is hidden and `bolton_record_id` returns an empty string with no indication of what went wrong. This makes debugging very difficult. Narrow the exception to `AttributeError` (the only realistic case for a missing relation on an unsaved/detached instance), or restructure with explicit `getattr` guards as done in `PhysicalRecord.bolton_record_id`.

bfd9000_web/archive/models.py
```diff
@@ -732,6 +848,62 @@ def subject(self) -> 'Subject':
...
+            self.sequence_number = (max_seq or 0) + 1
+        super().save(*args, **kwargs)
+        self._assign_bolton_record_identifier()
+
+    def _assign_bolton_record_identifier(self) -> None:
+        """Create or update the stored Bolton record identifier on the M2M identifiers field."""
+        new_value: str = self.bolton_record_id
+        if not new_value:
+            return
+        existing = list(self.identifiers.filter(system=SYSTEM_IDENTIFIER_BOLTON_RECORD))
```
zgypa: claude-sonnet-4.6 Medium: `SYSTEM_IDENTIFIER_BOLTON_RECORD` is shared between `PhysicalRecord` and `DigitalRecord`. Both models call `Identifier.objects.get_or_create(system=SYSTEM_IDENTIFIER_BOLTON_RECORD, value=new_value)` and then attach the result to their own `identifiers` M2M. A `PhysicalRecord` and its corresponding `DigitalRecord` will compute the same `bolton_record_id` string (same subject, record_type, age, sequence=1), causing `get_or_create` to return the *same* `Identifier` row, which then appears in both models’ identifier sets. This makes the identifier ambiguous — it is impossible to tell from the identifier alone whether it belongs to a physical or digital record. Either use separate system URIs (e.g. `bolton-physical-record-id` vs `bolton-digital-record-id`), or document explicitly that the sharing is intentional and that consumers must filter by model type.

bfd9000_web/archive/management/importers/richardson.py
```diff
@@ -0,0 +1,472 @@
...
+        if len(parts) < 3:
+            # Could be freeform text like "R Archive Box", or "10-9/10" (2 parts).
+            if len(parts) == 2:
+                # Format: "cabinet-slot1/slot2" with no shelf letter — infer shelf 'A'
+                cabinet = parts[0].strip()
+                slot_part = parts[1].strip()
+                slots = [s.strip() for s in slot_part.split('/')]
+                for slot in slots:
+                    if slot:
+                        loc, _ = PhysicalLocation.objects.get_or_create(
```
zgypa, line 413: claude-sonnet-4.6 Medium: `get_or_create` is keyed on `(cabinet, shelf, slot)` only, but `address` (the physical facility) is not part of the lookup. If two different facilities happen to use the same cabinet/shelf/slot numbering scheme, records from both will share the same `PhysicalLocation` row, silently pointing to whichever `address` was set first. Fix: include `address` in the lookup key, or — if all Richardson data belongs to a single known facility — always pass that facility’s `Address` in the `defaults` dict and note the assumption in a comment.

bfd9000_web/archive/templates/archive/base_index.html
```diff
@@ -728,7 +736,7 @@
 function attachSearchTooltip(containerId) {
```
zgypa, line 729: claude-sonnet-4.6 Low: The tooltip says searches use a "case-insensitive, substring match", but the actual `search_fields` on both `DigitalRecordViewSet` and `PhysicalRecordViewSet` are `['^identifiers__value']`. The `^` prefix in DRF SearchFilter means a **starts-with** (prefix) match, not a substring match. A user searching for "Lateral" or any mid-string fragment will get zero results, contrary to what the tooltip promises. Also the tooltip mentions searching by record type code/name, but `identifiers__value` does not cover record type codes or names at all. Either fix the tooltip to accurately describe the starts-with behaviour, or remove the `^` prefix from `search_fields` to get the substring match the tooltip advertises.

zgypa:
Applied changes suggested.

PR #94, opened 2026-05-30
Add strict type checking and move python package manager to uv

tnichols217:
Add strict type checking across the entire bfd9000_web codebase. The following tools are currently being used:

- Ruff - style, documentation, cleanliness, complexity
- Basedpyright - Strict LSP and code-time type checking
- mypy - Very strict type checking for Django to be used in CI

The environment setup has been significantly simplified due to using UV, which will also help in CI later down the line. The .envrc and flake no longer have complicated package management scripts and instead rely on UV's lockfile. The plan is to use this to build the docker container as well by the end of this PR. 

This PR is not yet complete, so I will be making it a draft. At the time of opening this PR, the app does run without errors, ruff reports 0 errors and 4 warnings, and mypy reports 7 warnings, down from over 15000 across basedpyright, ruff, and mypy before this refactoring.

Steps before this PR is complete:

- get the dockerfile fixed
- build testing pipelines to ensure code stays clean
- modernize nix builds (with tests)


tnichols217:
All listed steps above are now complete, I forgot to update this PR as i was working on CAS.

The dockerimage is now nix-built and tested as such, the repo now runs all tests as part of github actions, including:

1. The Django tests existing prior
2. Format checks
3. Type checking
4. Ruff lint checking

The github action for publishing the docker image is also now updated, utilizing a layered build strategy to deduplicate dependencies upon deployment, leaving only the actual code changing when performing new deployments. All django tests have passed as of this moment, there are still some minor format/mypy/ruff checks that do not pass but can be worked on in the future as CAS is the current priority.

tnichols217:
As seen in the github action, the only failing tests are mypy, ruff, and formatting

tnichols217:
This should be the final commit for this PR, adding a convenience function for building and loading the relevant built docker image directly into podman, making podman/docker compose utilize the latest built image after calling the provided `load-podman` function provided as part of the shell. All LSP issues have been fixed and works with the new immutable nix venv provided by uv2nix in a prior commit. As in previous commits as well, the development and production venvs are made available within the flake, which eases deployment and is used in the docker image builder. This is confirmed to work with the docker compose and all items build correctly. Running the formatter and fixing the remaining lint/type errors can be part of this PR or moved to a separate one.

zgypa-agent (changes requested):
copilot:claude-sonnet-4.6

The single CI run on this PR (`Run All Tests`, run 26884656573) **failed** with three broken checks. The PR description acknowledges outstanding lint/type issues and lists fixing them as a pre-completion step, so these are blocking.

**Failing checks:**

1. **`mypy-types`** — 7 errors in 4 files:
   - `archive/models.py:170,1292` — `Cannot determine type of "AUTH_USER_MODEL"` (`[has-type]`)
   - `archive/admin.py:352,427` — `"Callable[...]" has no attribute "short_description"` (`[attr-defined]`)
   - `archive/serializers.py:417` — Incompatible return type (`[return-value]`)
   - `archive/views.py:265` — `perform_create` argument incompatible with supertype (`[override]`)
   - `archive/views.py:377` — `filter_queryset` signature incompatible with supertype (`[override]`)

2. **`ruff-lint`** — 5 errors in `archive/serializers.py:1005` (`DigitalRecordSerializer.create`):
   - `PLR0912` Too many branches
   - `PLR0915` Too many statements (76 > 50)
   - `PLR0914` Too many local variables (30/15)
   *(A `# TODO: refactor this` comment is already present — these violations need either suppression with a tracking comment, or the refactor needs to happen here.)*

3. **`formatting`** (treefmt) — uncommitted auto-format diffs in:
   - `nix/bfd9000-docker.nix` (Nix formatting style)
   - `ruff.toml` (trailing comma in array)

Please fix (or explicitly suppress with justification) all three before removing draft status / requesting re-review.

bfd9000_web/BFD9000/conf.py
```diff
@@ -0,0 +1,15 @@
...
+but offers good type support otherwise
+"""
+
+from typing import TYPE_CHECKING
+
+from django.conf import (
+    settings as untyped_settings,  # pyright: ignore[reportUnusedImport]
+)
+
+import BFD9000.settings as typed_settings
```
zgypa-agent: copilot:claude-sonnet-4.6

`import BFD9000.settings as typed_settings` runs unconditionally at module load time (it is outside `TYPE_CHECKING`). At runtime `settings` resolves to `untyped_settings`, but the import still executes all of `settings.py`'s side effects a second time (including `monkeypatch()`). More importantly, this pattern now lives inside `0001_initial.py` (migration history), so it cannot be easily changed later.

The mypy `[has-type]` errors for `AUTH_USER_MODEL` in `models.py:170,1292` (CI run 26884656573) also indicate this approach is not fully solving the typing problem it targets.

If the intent is to follow the standard Django typed-settings pattern and the double import is considered acceptable, note that the import graph constraint is now implicit and easy to violate: any module that `settings.py` imports (directly or transitively) must not import `conf.py`, or you get a circular import at startup. It also means `conf.py` cannot be imported before Django is fully initialized (e.g. in standalone scripts or test fixtures that configure settings manually).

A simpler alternative: `django-stubs` already ships a typed stub for `settings.AUTH_USER_MODEL` — it works out-of-the-box when `django_settings_module` is set in `pyproject.toml` (which it is). The `[has-type]` mypy errors suggest the `conf.py` indirection is actually obscuring the type rather than exposing it. Dropping `conf.py` entirely and using `from django.conf import settings` uniformly would resolve both the circular-import risk and the mypy errors, with `TYPE_CHECKING`-only annotations only for app-specific attributes (`APP_VERSION`, `FORCE_SCRIPT_NAME`) that are not in the standard stubs.

.github/workflows/publish-ghcr.yml
```diff
@@ -60,27 +56,19 @@ jobs:
...
-          file: ./bfd9000_web/Dockerfile
-          push: true
-          tags: ${{ steps.meta.outputs.tags }}
-          labels: ${{ steps.meta.outputs.labels }}
-          cache-from: type=local,src=/tmp/.buildx-cache
-          cache-to: type=local,dest=/tmp/.buildx-cache,mode=max
+      - name: Tag and Push Image to Registry
+        run: |
+          # Bind local build to remote metadata tag string
+          REMOTE_TAG="${{ steps.meta.outputs.tags }}"
```
zgypa-agent, line 57: copilot:claude-sonnet-4.6

`steps.meta.outputs.tags` is a newline-separated string that can contain multiple tags. Assigning it whole to `REMOTE_TAG` and passing it to `docker tag` / `docker push` works only when exactly one tag is produced. With the current `type=semver,pattern={{version}}` config that holds — but the comment that previously documented this single-tag constraint was removed in this PR, making the assumption invisible.

If a second entry is ever added to the `tags:` block, both `docker tag` and `docker push` will receive a multi-line argument and fail. Consider either:

```yaml
# Option A: explicit loop
- name: Tag and Push Image to Registry
  run: |
    while IFS= read -r tag; do
      [ -z "$tag" ] && continue
      docker tag edu.case.bfd9000:latest "$tag"
      docker push "$tag"
    done <<< "${{ steps.meta.outputs.tags }}"
```

or restore a comment making the single-tag assumption explicit.

bfd9000_web/pyproject.toml
```diff
@@ -0,0 +1,75 @@
...
+version = "0.1.0"
+description = "Add your description here"
+readme = "README.md"
+requires-python = ">=3.13"
+dependencies = [
+    "cryptography>=44.0,<45.0",
+    "django>=6.0,<7.0",
+    "django-cors-headers>=4.0,<5.0",
+    "django-filter>=25.0,<26.0",
+    "django-stubs>=6.0,<7.0",
```
zgypa-agent: copilot:claude-sonnet-4.6

`django-stubs` is listed under `[project.dependencies]` (production), which means it is installed in the production Docker image and in every deployment. It is a type-stubs-only package (~10 MB of `.pyi` files) with no runtime behaviour beyond the `monkeypatch()` call in `settings.py`.

`monkeypatch()` itself is safe to call at runtime, but shipping the full `django-stubs` package to production adds unnecessary size. AGENTS.md previously noted `django-types` as a dev-time dependency — the same principle applies here.

Consider moving `django-stubs` (and `djangorestframework-stubs`) to `[dependency-groups.dev]` only, and guarding the `settings.py` import so production startup does not fail if the package is absent:

```python
try:
    from django_stubs_ext import monkeypatch
    monkeypatch()
except ImportError:
    pass
```

zgypa-agent:
copilot:claude-sonnet-4.6

`AGENTS.md` still describes the old `requirements.txt`-based workflow as the authoritative source of truth for Python dependencies (lines 5–7), but both `requirements.txt` and `requirements-dev.txt` were deleted in this PR. The `.envrc` now contains only `use flake`, and dependencies are managed via `pyproject.toml` + `uv.lock`.

Because `AGENTS.md` is read by AI coding agents at every session start, leaving it pointing to deleted files will cause every AI-assisted contribution to follow the wrong workflow until this is fixed. This is a one-paragraph update — worth doing in this PR rather than as a follow-up.

Suggested replacement for the affected paragraph:

> Dependencies are managed via `bfd9000_web/pyproject.toml` and the `uv.lock` lockfile — these are the **only** source of truth for Python dependencies. To add or update a dependency, edit `pyproject.toml` and run `uv lock` (or `uv add <pkg>`) inside `bfd9000_web/`. Do not add Python packages to `flake.nix`. The dev shell (`nix develop`) provides `uv` and the full dev venv automatically via `uv2nix`.

zgypa-agent:
copilot:claude-sonnet-4.6

This PR introduces several new tools that are not documented anywhere a developer or AI agent would find them:

- **ruff** — `ruff.toml` is present and enforced in CI (`nix flake check`), but no doc explains what rule sets are selected, how to run it locally, or — critically — that it replaces pylint. Without this, another developer encountering `.pylintrc` at the repo root will reasonably assume pylint is still the active linter and may re-introduce it or run it alongside ruff.
- **uv** — `pyproject.toml` + `uv.lock` are now the dependency source of truth, but nothing explains how to add/update a dependency (`uv add`, `uv lock`), why it replaced `pip`+`requirements.txt`, or that `requirements.txt`/`requirements-dev.txt` are intentionally gone. Without this, another developer will recreate those files or reach for `pip install` when adding a package.
- **mypy strict mode** — enforced in CI via `nix flake check`, but not mentioned in any developer-facing doc.
- **basedpyright** — mentioned in the PR description as the LSP tool, but not documented.

Meanwhile several stale references remain:
- `AGENTS.md` Typing section still references `django-types` (removed in this PR, replaced by `django-stubs`) and `VSCode + Pylance` without mentioning basedpyright.
- `bfd9000_web/README.md` "Local Docker Workflow" section still describes `docker build -f bfd9000_web/Dockerfile` — the Dockerfile was deleted in this PR.

Suggested additions (can be in this PR or a dedicated follow-up, but should land before merging to `develop`):

1. **`AGENTS.md`**: Update the Typing section to reflect `django-stubs`, `mypy`, and `ruff` as the actual toolchain. Add a short "Code Quality" paragraph listing each tool, its role (lint / type-check / format), and the local invocation:
   - `ruff check ./bfd9000_web` — lint (replaces pylint; `.pylintrc` is superseded)
   - `ruff format ./bfd9000_web` — format
   - `mypy` (from inside `bfd9000_web/`) — strict type-check
   - `nix flake check --keep-going` — full CI gate locally

   Add a "Dependency management" paragraph explaining that uv replaced pip+requirements.txt, that `pyproject.toml` and `uv.lock` are the only source of truth, and that `uv add <pkg>` / `uv lock` are the correct commands — so no developer recreates the deleted files.

2. **`bfd9000_web/README.md`**: Add a "Running linters locally" section with the above commands, and update the dependency section to explain `uv add <pkg>` / `uv lock` as the way to manage dependencies.

zgypa-agent:
copilot:claude-sonnet-4.6

The `Dockerfile` was deleted in this PR, but several parts of the repo still depend on a buildable image and have not been updated:

- **`docker-compose.yml`** references `image: localhost/edu.case.bfd9000:latest` with no `build:` context. `docker compose up` will fail on any machine where that image does not already exist locally.
- **`Makefile`** `build-test-docker` calls `docker compose -f $(COMPOSE_FILE) build` — this will error with no Dockerfile present.
- **`DEPLOY.md`** lists `make build-test-docker` and `make run-test-docker` as release steps — both are now broken.
- **`bfd9000_web/README.md`** "Local Docker Workflow" section describes `docker build -f bfd9000_web/Dockerfile` — the file no longer exists.

The new build path appears to be `nix build .#dockerImage` + `load-podman` (from the nix dev shell), but this is not documented anywhere. The production server orchestrates with docker-compose and does not necessarily have Nix available.

Before this merges to `develop`, the following needs to be resolved:

1. How does a developer (or the production server) obtain the image for `docker compose up`? There are a few options:
   - Restore a `Dockerfile` alongside the Nix build for environments without Nix.
   - Document and script the `nix build .#dockerImage` → `docker load` step as the replacement for `make build-test-docker`, and update `Makefile` accordingly.
   - Document the minimal Nix installation required on developer and server machines (`nix` single-user install is sufficient), so the Nix-based build is a first-class supported path everywhere.

2. Whichever option is chosen, `Makefile`, `DEPLOY.md`, and `bfd9000_web/README.md` must be updated to reflect the new workflow.

zgypa (changes requested):
Seems like @zgypa-agent found a number of issues to clean up for this cleanup PR. I can pass you the prompt/skill i use for doing PRs, if you wish, so you can run it yourself, locally, and save a couple of PR review roundtrips.

tnichols217:
Ill work on fixing all this this week, most of it is just documentation, which I havent done yet for this PR, so it makes sense. It seems theres no structural issues with the code though, which is good. Documentation won't take too long to write. Ill make my way through the documentation and be sure to update all of it

zgypa:
> Ill work on fixing all this this week, most of it is just documentation, which I havent done yet for this PR, so it makes sense. It seems theres no structural issues with the code though, which is good. Documentation won't take too long to write. Ill make my way through the documentation and be sure to update all of it

@tnichols217  OK: i removed the review requests then. You just changed this to draft status, there is nothing for me to review yet.

tnichols217:
It should be good now, all checks should be passing and all items are addressed in the latest commit

Makefile
```diff
@@ -9,16 +9,17 @@ help: ## Show available targets
 	@awk 'BEGIN {FS = ":.*##"} /^[a-zA-Z0-9_-]+:.*##/ {printf "%-20s %s\n", $$1, $$2}' $(MAKEFILE_LIST)
 
 build-test-docker: ## Build test Docker image with compose
-	docker compose -f $(COMPOSE_FILE) build
+    podman-load
```
zgypa-agent: claude:sonnet-4-6

`Makefile:12` (`build-test-docker` target) is broken:

1. The new `podman-load` line is indented with spaces instead of a tab. GNU Make requires recipe lines to start with a tab, so this causes a fatal `Makefile:12: *** missing separator. Stop.` error for **any** `make` invocation, not just this target.
2. `podman-load` isn't a defined command — the utility this PR documents elsewhere (`bfd9000_web/README.md`, `nix/load-podman.nix`) is `load-podman`.

Suggest either fixing both (tab + correct command name) or removing the line if it wasn't intentional.

bfd9000_web/README.md
```diff
@@ -9,7 +9,7 @@ To use it, install [Nix](https://nixos.org/download.html) and run the following
 nix develop
```
zgypa-agent, line 10: claude:sonnet-4-6

`bfd9000_web/README.md:102` (Local Docker Workflow) instructs `cp .env.example .env`, but the actual file is `dot-env.example`. Following the README verbatim fails with "cp: cannot stat .env.example: No such file or directory".

Please update to `cp dot-env.example .env`, or rename the source file if that was the intent.

flake.nix
```diff
@@ -2,84 +2,161 @@
...
+          pythonEnv = venv;
+        };
+        load-podman = pkgs.callPackage ./nix/load-podman.nix { };
+        treefmtconfig = inputs.treefmt-nix.lib.evalModule pkgs-treefmt {
+          projectRootFile = "flake.nix";
+          programs = {
+            alejandra.enable = true;
+            toml-sort.enable = true;
+            yamlfmt.enable = true;
+            # mdformat.enable = true;
```
zgypa-agent: claude:sonnet-4-6

`README.md` (Code Cleanliness section) says formatters cover "markdown, yaml, toml, or sh" via treefmt.

But `flake.nix:54` was changed in this PR to comment out `mdformat.enable = true;`. Markdown formatting is no longer enforced — the CI `formatting` check now only runs alejandra, toml-sort, yamlfmt, shellcheck, shfmt, and nixfmt.

The README overstates formatter coverage. Either remove "markdown" from the list, or re-enable mdformat.

zgypa-agent:
claude:sonnet-4-6

Thanks for the doc pass — a good chunk of the prior feedback is addressed (AGENTS.md dependency workflow, README "Code Cleanliness" section, Local Docker Workflow rewrite). However, not everything is addressed yet at `b7c5f20`:

Still open from the prior round:
- `bfd9000_web/BFD9000/conf.py:13` — conf.py unchanged, still unconditionally imports `BFD9000.settings` at runtime (see thread).
- `.github/workflows/publish-ghcr.yml:57` — `REMOTE_TAG` single-tag assumption still undocumented, code unchanged (see thread).
- `bfd9000_web/pyproject.toml` — `django-stubs`/`djangorestframework-stubs` are still listed under `[project.dependencies]` (production), not just `[dependency-groups.dev]` (see thread).

Two doc-only gaps not yet covered by any thread:
- `.pylintrc` is still present at the repo root with no doc noting that ruff has superseded it — a future contributor could reasonably reintroduce pylint.
- `DEPLOY.md` (lines 7-8) still lists `make build-test-docker` / `make run-test-docker` as release steps; both are currently broken (see the new Makefile:12 thread).

Also flagged today (new issues introduced by this doc pass): `Makefile:12` (broken `build-test-docker` target), `bfd9000_web/README.md:102` (`.env.example` doesn't exist — actual file is `dot-env.example`), and `flake.nix:54` (`mdformat` disabled but README still claims markdown formatting).

Please reply on each open inline thread (the three above plus the three new ones from today) describing what was changed or why it's not applicable — leave the threads unresolved, the reviewer will resolve them after verifying.

zgypa-agent (changes requested):
Forgot to publish this review. There are some findings from 4-5 days ago.

zgypa (commented):
So there are a bunch of findings from the agent, which i forgot to publish... Sorry.

Take a look at them, see if they make sense. They seemed to make sense when i went through this a few days ago, but i can't remember now. I do think some are relevant.

Add your replies/changes, and leave unresolved.

Thanks!

tnichols217:
> claude:sonnet-4-6
> 
> Thanks for the doc pass — a good chunk of the prior feedback is addressed (AGENTS.md dependency workflow, README "Code Cleanliness" section, Local Docker Workflow rewrite). However, not everything is addressed yet at `b7c5f20`:
> 
> Still open from the prior round:
> 
>     * `bfd9000_web/BFD9000/conf.py:13` — conf.py unchanged, still unconditionally imports `BFD9000.settings` at runtime (see thread).
> 
>     * `.github/workflows/publish-ghcr.yml:57` — `REMOTE_TAG` single-tag assumption still undocumented, code unchanged (see thread).
> 
>     * `bfd9000_web/pyproject.toml` — `django-stubs`/`djangorestframework-stubs` are still listed under `[project.dependencies]` (production), not just `[dependency-groups.dev]` (see thread).
> 
> 
> Two doc-only gaps not yet covered by any thread:
> 
>     * `.pylintrc` is still present at the repo root with no doc noting that ruff has superseded it — a future contributor could reasonably reintroduce pylint.
> 
>     * `DEPLOY.md` (lines 7-8) still lists `make build-test-docker` / `make run-test-docker` as release steps; both are currently broken (see the new Makefile:12 thread).
> 
> 
> Also flagged today (new issues introduced by this doc pass): `Makefile:12` (broken `build-test-docker` target), `bfd9000_web/README.md:102` (`.env.example` doesn't exist — actual file is `dot-env.example`), and `flake.nix:54` (`mdformat` disabled but README still claims markdown formatting).
> 
> Please reply on each open inline thread (the three above plus the three new ones from today) describing what was changed or why it's not applicable — leave the threads unresolved, the reviewer will resolve them after verifying.

1. conf has bee updated
2. fixed
3. django-stubs is required at runtime in order to monkeypatch() properly, we cannot simply "pass" if it cannot import as the type resolutions at runtime will fail and crash at unexpected times. monkeypatch exists to make sure we can safely operate the django framwork in a type-safe manner. The second point is just false, the stubs are not part of production.
4. removed
5. fixed
6. fixed
7. mdformat will remain off until Toni and I have settled on an md formatting standard to follow, documentation can remain as assuming markdown will be formatted as this will happen soon.

tnichols217:
@zgypa all of them were trivial issues, it should all be completed now.

Makefile
zgypa-agent: claude:sonnet-4-6

Thanks for the fix — the command name is now correct (`load-podman`), but line 12 is still indented with 4 spaces instead of a tab. GNU Make requires recipe lines to start with a tab, so `make` still fails immediately on *any* target:

```
$ make -n build-test-docker
Makefile:12: *** missing separator.  Stop.
```

This breaks every `make` invocation, not just `build-test-docker`. Please replace the leading spaces on line 12 with a single tab character to complete the fix.

This is also why `DEPLOY.md:7-8` (`make build-test-docker` / `make run-test-docker`) still don't work — those lines are correct as written and need no edit; they'll work again once this indentation is fixed.

zgypa-agent (changes requested):
claude:sonnet-4-6

## Final review pass before merge (HEAD `021ddc7`)

**CI**: now green — `mypy-types`, `ruff-lint`, and `formatting` (treefmt) all pass via the nix flake checks. The previously-failing `PLR0912`/`PLR0914`/`PLR0915` complexity warnings on `DigitalRecordSerializer.create` (`serializers.py:999`) are addressed via a documented `# noqa: C901, PLR0912, PLR0914, PLR0915` plus the existing `# TODO: refactor this, it's way too complex and prone to erroring` — consistent with the "suppress with justification" option from the earlier review. Not blocking.

**Blocking**:
- `Makefile:12` (`build-test-docker`) — see the open thread. The command name was fixed (`load-podman`), but the recipe line is still space-indented instead of tab-indented, so `make` fails immediately with `Makefile:12: *** missing separator. Stop.` for *every* target, not just this one. This also blocks `DEPLOY.md`'s release checklist (`make build-test-docker` / `make run-test-docker`). One-character fix: replace the leading spaces on line 12 with a tab.

**Non-blocking suggestion**:
- `bfd9000_web/BFD9000/settings.py` — `APP_VERSION` lost its `try/except OSError` fallback when it was rewritten as an `or`-chain:
  ```python
  APP_VERSION = (
      os.environ.get("APP_VERSION")
      or APP_VERSION_FILE.read_text(encoding="utf-8").strip()
      or "nover"
  )
  ```
  If `bfd9000_web/VERSION` is ever missing and `APP_VERSION` isn't set via env, this now raises `FileNotFoundError` at settings-import time (hard Django startup crash) instead of falling back to `"nover"`. Low risk today — `VERSION` is tracked in git and explicitly copied by `nix/bfd9000-app.nix`, and CI passes — but worth restoring the fallback if you'd rather degrade gracefully than crash in an unforeseen build context.

Once `Makefile:12` is fixed, this looks good to merge from my side.

zgypa (changes requested):
I confirmed myself that, indeed, running `make` is still broken. Which is surprising to me: how did you test building, then? If you have a new build path, and Makefile is not needed anymore, then maybe we should move to that, and remove Makefile, and adjust documentation accordingly. Otherwise, simply fix the Makefile, and you should be done.

tnichols217:
i've personally been running the commands directly in my terminal, I was not aware of the Makefile until the PR informed me it needed to be updated. I had just been building/running with the script I've provided, and manually managing the compose by command-line, as it provides me more flexibility in testing. Now that I know theres a makefile, I'll be using that instead though.

tnichols217:
as for the app version issue, we can cleanly resolve that with a result in a later commit, as result is added with the box PR. All other issues have been solved at this point, I dont think there are any blockers remaining

flake.nix
zgypa-agent: copilot:claude-haiku-4-5-20251001

The README claim on line 35 is factually incorrect: markdown formatting is not currently enabled. Looking at `flake.nix` line 54, `mdformat.enable = true;` is commented out.

**Recommendation:** Remove "markdown" from the formatter description in README.md:35 (change "markdown, yaml, toml, or sh" to just "yaml, toml, or sh") until mdformat is re-enabled, to keep the documentation accurate with the actual tooling configuration.

Makefile
zgypa-agent: copilot:claude-sonnet-4.6

Tab indentation confirmed at HEAD (`2d7fff5`). `make build-test-docker` and `make run-test-docker` are correctly tab-indented and use `load-podman` + `podman-compose`. Resolving.

tnichols217:
Markdown formatting is specified within the `flake.nix`, with 88 width, no numbering, LF line endings, and to be github markdown compliant (as we are using GitHub). It additionally vertically aligns tables and standardizes the `  ` line endings as `\` instead, which is visible in diffs and editors. If you don't like this change, we can modify the settings or remove markdown formatting altogether. If you don't have an editor open to see the new md files, here is a screenshot of what the file looks like when being viewed:

image


as well as a table:

image

tnichols217:
As with all the other formatters, linters, and checkers, it is integrated and can be run with the `nix fmt` command and is also checked during CI

bfd9000_web/pyproject.toml
zgypa-agent: copilot:claude-sonnet-4.6

Thanks for the explanation — keeping `monkeypatch()` as a production runtime call is reasonable if any runtime code subscripts Django classes outside `TYPE_CHECKING` guards.

However, the factual claim in your June 11 reply ("the stubs are not part of production") does not match the code at HEAD (`2d7fff5`): `django-stubs` is still listed in `[project.dependencies]` at `pyproject.toml:35` (the production list). It also appears at line 7 in `[dependency-groups].dev`, so it is in both.

Two possible resolutions:
1. **Keep it in production** (if `monkeypatch()` is intentionally a production runtime requirement) — remove line 35 is not needed, but please confirm this intent here so the thread can be closed.
2. **Move to dev-only** — remove line 35 from `[project.dependencies]` and guard `settings.py`:
```python
try:
    from django_stubs_ext import monkeypatch
    monkeypatch()
except ImportError:
    pass
```

Either way, the current state (in both lists) is likely unintentional.

flake.nix
zgypa-agent: copilot:claude-sonnet-4.6

Confirmed at HEAD (`b8540d1`) — mdformat is now enabled in `flake.nix:54` with GFM plugin, 88-char wrap, and LF endings. The README claim is now accurate. Resolving.

bfd9000_web/pyproject.toml
zgypa-agent: copilot:claude-sonnet-4.6

Correction noted — `djangorestframework-stubs` is correctly dev-only (line 9 only), that point was already addressed. Apologies for the misread.

The remaining question is `django-stubs` at line 35 in `[project.dependencies]`. Given your explanation that `monkeypatch()` is an intentional production runtime call, keeping `django-stubs` in production is consistent with that intent. The duplication (it also appears in the dev group at line 7) is redundant since the production dep covers all environments, but it is not a bug.

If the intent to keep `django-stubs` in production for `monkeypatch()` is confirmed, I'll consider this thread resolved — you can resolve it.

zgypa-agent (approved):
LGTM

zgypa (approved):
LGTM

tnichols217:
squash merge?

PR #95, opened 2026-06-03
Add basic CAS support

tnichols217:
Implements basic CAS support by using django-cas-ng. New dependency has been added to the uv lock. Tests have not been written for this module yet, still not sure how we want to implement that - possibly by having a mock CAS server to attempt logins with or just leave that portion untested as it is primarily dealt with by an external dependency anyway? 

Caddy has been added to the docker compose as the CAS server requires redirecting to an https address (even if it's localhost), additionally new environment configuration variables are made available:

CAS_ENDPOINT and PROXIED_URL, which represent which CAS server we are querying against, exposed to enable testing down the line, and the url to request the CAS server to redirect to after authentication.

The design of the login page is also non-final and seeking feedback (as I am not a UI designer), a screenshot is attached below:

image


zgypa (commented):
- Don't test functionality of external deps. Assume they work. Only test within repo boundaries. 
- Once deployed, they system should be using a static .case.edu SSL cert, so we do have https:// access, using the django's gateway thing web server. Why do we need an extra web server?
- Design is fine, don't worry too much about it, just make sure you use vanilla setups, w/o adding any extra complexity, s.t. future modification/improvement if we pass to a designer, will be trivial.
- Do we even need username/password setup? I would actually directly re-direct to SSO login. Then we can have a custom path, that we know about, that we use for local user/pass login, in case CAS breaks. Like bfd9000/login/ or something. While the default takes the user straight to CAS login, no questions asked. KISS it.

tnichols217:
the proxy is just for local development, as the SSO will refuse to redirect if the target is not https. Since we have to redirect to case's login page, and they voluntarily redirect back to our page, but will refuse if its not https

tnichols217:
im going to prep this branch for merging soon, since other features are nearing completion. I've thought about the removal of email/password, but I think it might be best to have it exist as a fallback. The feature to have both is already implemented and it wouldn't be much simpler to remove the email/password at this point. It would also serve as a bonus for allowing us to maintain the system if case ever changes their SSO tactic and we fail to resolve with them.

bfd9000_web/archive/templates/archive/login.html
```diff
@@ -35,6 +35,17 @@ <h1 class="text-3xl font-bold mb-2">BFD9000</h1>
             {% endif %}
         {% endif %}
 
+        <div class="cas-login">
+            <a href="{% url 'cas_ng_login' %}" class="btn btn-primary w-full mt-4 flex items-center justify-center text-center">
```
zgypa-agent: claude:sonnet-4.6
Question based on reading the code (haven’t tested this flow end-to-end): the "Sign In with SSO" link just points to `{% url 'cas_ng_login' %}` with no `next` param, while the local login form below it forwards `next` via a hidden input (line 75). If a user lands here via `?next=/some/page/` and clicks "Sign In with SSO", does django-cas-ng still send them back to that page afterward, or would it fall back to `CAS_REDIRECT_URL`/the index? If the latter, might be worth appending `?next={{ next }}` to the SSO link.

zgypa-agent:
claude:sonnet-4.6
Makes sense — CAS requires an HTTPS service URL to redirect back to, and Caddyfile's `local_certs` + the new `caddy` service is a minimal, vanilla way to get that locally without extra deps (matches the `SECURE_PROXY_SSL_HEADER`/`CAS_ROOT_PROXIED_AS` setup in settings.py). One follow-up: `CAS_ENDPOINT` and `PROXIED_URL` aren't documented in `dot-env.example` or the README, and there's nothing telling a new dev they need to hit `https://localhost:4430` (not :8000) to exercise the CAS login flow locally. Worth adding a short note?

zgypa-agent:
claude:sonnet-4.6
Reconsidering with the goal of minimizing user confusion (most users won't read a choice between SSO and password) — would it make sense to have `/login/` redirect straight to CAS by default (e.g. via `LOGIN_URL`), and move the existing password form to a separate, documented path like `/login/local/` for admins/support only? Django's `login_required` already appends `?next=` to `LOGIN_URL`, and django-cas-ng's `LoginView` reads that same param, so this might also resolve the `next`-forwarding question raised on the other thread on this page.

Separately — and likely out of scope here, since this PR is specifically about CAS — there's a related UX gap worth its own issue: if a CAS-authenticated user doesn't have access to the app (either because their account genuinely lacks permissions, or because they logged in with a shared/wrong credential), they should get a clear "you don't have access" message with a link/instructions on how to request access, plus an obvious way to log out and try a different account. These feel like the same underlying no-access UX concern and could be tracked together in a follow-up issue. Happy to move forward with this PR as-is and have that tracked separately.

zgypa (commented):
Good job, can't wait to try it out!

I'm also concerned about the confusion: the user sees a user/pass prompts, and naturally tries to use that. I also suggest just simply moving user/pass to another URL, that only we know, and is documented, possibly in Canvas also. 

Does that makes sense?

zgypa-agent (commented):
claude:sonnet-4.6
First-round review pass: one open inline question (next-redirect handling on the SSO login link) below, plus two general comments posted above on the local-dev proxy rationale and the CAS-first login UX discussion.

tnichols217:
The login page now looks like this:

image

I decided to make you still click to do the SSO as without it, there is no page nor confirmation that the login actually happened. In the future, if we decide to access control the SSO to specific people, an error can be displayed on this page as well. Pressing the SSO button will immediately log you in with SSO, if you are already logged into Case on that device, there is no confirmation page from Case's side, so its instant.

zgypa-agent (commented):
Deleted

zgypa-agent:
model:claude-sonnet-4-6

The README table now documents both variables (lines 110–111), but `dot-env.example` hasn't been updated yet. Please add `CAS_ENDPOINT` and `PROXIED_URL` to `dot-env.example` with their default example values so a developer copying that file gets the right starting point.

A brief note in README about needing HTTPS on port 4430 (not :8000) for the local CAS flow would also help new developers avoid that friction.

zgypa (approved):
No further PR necessary as long as you implement the two minor adjustments commented by the agent. Then you can merge.

Thanks!

zgypa:
Feedback on the current login approach:

1. **Remove the choice FROM the page** — `/login/` should only have the Button. There is no reason to present users with a choice, because the users get confused, and will inevitebly click on the wrong login option. The local login path is for internal/admin use only and must not be discoverable from the UI.

2. **Button label**: Use **"Log in with Case ID"** — because users have no idea what "SSO" means, and some will not want to click on something they don't know, and will call us.

3. **Local login path**: Rename to `/login/local/` (not `/login/email/`). Because i don't find email appropriate name here. local is more correct.

tnichols217:
everything should be up to date now. the dot-env.example was not updates as I explicitly marked them as non-mandatory in the README, following the trend with the other env variable in the example, for only being present if marked as required in the README.

tnichols217:
Closed #18

zgypa-agent (changes requested):
openai:gpt-5.5 I left two findings: one line-anchored version regression where the PR downgrades `bfd9000_web/VERSION` from `0.2.1-dev` to `0.2.0`, and one PR-wide documentation gap for the new local Caddy/CAS flow. The version regression affects app/API-visible release metadata, so I am requesting changes until that is corrected.

zgypa-agent:
openai:gpt-5.5 Correction: my `REQUEST_CHANGES` review at https://github.com/open-ortho/edu.case.BFD9000/pull/95#pullrequestreview-4585230334 was posted to PR #95 by mistake while reviewing PR #96. I deleted the accidental inline and issue comments from this PR. GitHub is not changing/removing the submitted review state on this already-merged PR, so please disregard that review. The VERSION finding does not apply to PR #96, and the still-relevant local CAS/Caddy docs finding has been moved to PR #96 here: https://github.com/open-ortho/edu.case.BFD9000/pull/96#issuecomment-4817740092

PR #96, opened 2026-06-06
Implement Box support and overhaul file support

tnichols217:
Initializing this PR now so that review can happen.

Merged from old Box development branch, still some fix up to, Storage backends are implemented, with an interface defined as well.

Ongoing efforts:

1. rewrite the sync thread or not do it at all
2. rewrite tests for the storage backend
3. integrate better with views/serializers, as the merge has them depending on the old interface
4. Clean up media utils and make file types consistently detected (by magic -> MIME)
5. Remove unused code from the previous merge
6. Update settings to use the new Box auth methods

Some large portions of code are comments as of the opening of this draft, this will either be migrated or removed by merge time.


tnichols217:
All efforts are complete, the sync thread was removed. All local tests are passing, but we haven't tested in staging yet to see if it works with real box or not. Once we get the box development token, that can be our next step before merging

zgypa:
@tnichols217 what's your plan with the other related PRs, that this PR would be making obsolete? should be linked somehow, so they don't get accidentally forgotten, or closed.

tnichols217:
All related work from related PRs were merged into this branch within the first commit. Merging this PR should allow us to properly close all other related PRs.

tnichols217:
Since the other branches have merged by now, I will rebase, squash, then continue testing on this branch before we can merge into develop

tnichols217:
box support is effectively complete and functional through tests. A couple minor things that could be added to this PR:

1. STLs dont have thumbnail previews
2. The digital record screen is slow because it queries every file in order to obtain its size, this is slow when box is the backend
3. This branch still needs to be merged with dev before I will request a full review
4. box files are NOT cached on the server, repeatedly refreshing a digital record page will send multiple box requests to fetch the file.

zgypa:
@tnichols217 : It should be pretty easy to get thumbnails from STLs... there's libs all over. Question is, in what orientation.
I agree caching is important, but add it as another separate issue, which we can tackle, when we need to improve performance.

tnichols217:
fixed all issues except for syncing with dev. Will sync now then I'll run tests and we should be able to merge

tnichols217:
All synced, ready for review.

tnichols217:
Closes #39 , #43

zgypa-agent:
openai:gpt-5.5 This branch still leaves the local CAS/Caddy flow under-documented. The compose file includes the Caddy HTTPS proxy on `https://localhost:4430`, and `PROXIED_URL`/`CAS_ROOT_PROXIED_AS` default to that origin, but `bfd9000_web/README.md` still tells developers to open `http://127.0.0.1:9000`, and the newly added `bfd9000_web/.env.example` does not include `CAS_ENDPOINT` or `PROXIED_URL`. A developer following the current docs or copying the new env example will bypass/miss the proxy settings that CAS callback testing depends on. Please update the local Docker workflow/env docs to describe when to use `https://localhost:4430`, how Caddy fits into compose, and what direct port `9000` access does or does not support.

bfd9000_web/archive/storage/storage.py
```diff
@@ -0,0 +1,361 @@
...
+        Args:
+            handle: The file handle to close
+
+        Returns:
+            True if closed, False if already closed
+
+        """
+
+    @abstractmethod
+    def delete(self, path: PathType) -> bool:
```
zgypa-agent, line 157: openai:gpt-5.5 Issue #43 originally scoped the storage layer to upload/list/download and said “NO delete/rename” because those operations should be done manually on the storage system. This interface now makes deletion a required backend operation, and the Django adapter plus Box backend wire it through to real Box deletes. If delete is necessary for Django storage compatibility or overwrite semantics, that is reasonable, but it changes the original safety model for archived originals. Please document the rationale and the intended guardrails for when app-driven deletes are allowed, so future code does not treat this as a general-purpose archive deletion API.

bfd9000_web/archive/storage/urihandler.py
```diff
@@ -0,0 +1,204 @@
...
+
+        Args:
+            handle: The handle to close
+
+        Returns:
+            True if closed successfully
+
+        """
+        session = self._session_registry[handle]
+        return self._routes[session.backend]._raw_close(session.handle)
```
zgypa-agent: openai:gpt-5.5 `_raw_close()` forwards the close to the routed backend but leaves the outer handle in `_session_registry`. Every Django file operation through `DjangoStorageAdapter` opens via this router and calls `close()`, so a long-running worker will retain one `_SessionEntry` per upload/download/size-generating operation for the lifetime of the process. Please pop the outer handle when the routed close succeeds (matching the concrete backends' behavior) so closed sessions are actually evicted.

bfd9000_web/archive/storage/box.py
```diff
@@ -0,0 +1,490 @@
...
+        Returns:
+            True if file existed and was deleted
+
+        """
+        client = self._box_auth.get_client()
+        parent_dir = self._resolve_path_to_id(client, path.parent, is_folder=True)
+
+        item = self._get_item(client, parent_dir.id, path.name, is_folder=False)
+
+        if not item:
```
zgypa-agent, line 270: openai:gpt-5.5 `_get_item()` is cached, and the overwrite path (after deleting a file for a big-file replace) correctly calls `cache_clear()`/pops the cache entry before continuing. `delete()` and `rmdir()` do not: they call `client.files.delete_file_by_id()` / `client.folders.delete_folder_by_id()` directly and return without touching the cache. A later `exists()` or other lookup for the same path can read the stale cached item and report a just-deleted file/folder as still present. I reproduced this directly — after a successful `delete()`, `exists()` still returned `True` from cache without another Box API call. Please invalidate the relevant `_get_item` cache entry on successful delete/rmdir, matching the pattern already used for the overwrite path.

bfd9000_web/archive/serializers.py
```diff
@@ -733,12 +749,37 @@ def get_thumbnail_url(self, obj: DigitalRecord) -> str | None:  # noqa: PLR6301
...
+                )
+                obj.thumbnail = refreshed_obj.thumbnail
+
+        try:
+            return obj.thumbnail.url
+        except Exception as e:
+            logger.warning(f"Thumbnail fetch failed with {e}")
+            return None
 
     def get_image_url(self, obj: DigitalRecord) -> str | None:  # noqa: PLR6301
```
zgypa-agent, line 784: openai:gpt-5.5 `api_requirements.md` requires `image_url`/`thumbnail_url` to be relative API paths (e.g. `/api/records/123/image/`), and this PR actually re-touches that exact "Image URLs" section (rewrapping it) without changing the requirement. However, `get_image_url()`/`get_thumbnail_url()` return `obj.source_file.url`/`obj.thumbnail.url` directly, and with the new `DjangoStorageAdapter`, `.url()` just echoes the stored name verbatim (e.g. `box://<uuid>.png`) instead of an API path — confirmed directly: `settings.GET_DJANGO_STORAGE().url("box://abc.png")` returns `"box://abc.png"`. This lines up with "integrate better with views/serializers, as the merge has them depending on the old interface" in the PR description's ongoing-efforts list, so this may just be one of the known not-yet-done items rather than something missed. If so, could you confirm and track it before merge (or here) so it isn't lost? If it's actually intentional for some reason, the spec should be updated to match.

bfd9000_web/archive/serializers.py
```diff
@@ -733,12 +749,37 @@ def get_thumbnail_url(self, obj: DigitalRecord) -> str | None:  # noqa: PLR6301
...
+            if not thumb_bytes:
+                logger.warning("No thumbnail made")
                 return None
-        return None
+
+            with transaction.atomic():
+                refreshed_obj = DigitalRecord.objects.select_for_update().get(id=obj.id)
+                logger.info(f"Regeneration successful for {refreshed_obj.id}")
+                refreshed_obj.thumbnail.save(  # pyright: ignore[reportUnknownMemberType]
+                    f"{name}.jpg", ContentFile(thumb_bytes), save=True
```
zgypa-agent: openai:gpt-5.5 `_render_thumbnail_from_raster()` explicitly encodes thumbnail bytes as WebP (`processed.save(out, format="WEBP", ...)` in `media_utils.py`), but both places that persist the result — the regeneration path here (`f"{name}.jpg"`) and the upload path in `DigitalRecordUploadSerializer` (`f"{file_uuid}.jpg"`) — save it with a `.jpg` extension. The stored file's content/extension mismatch can break clients, CDNs, or image libraries that determine format from the filename rather than sniffing bytes. Please save thumbnails with a `.webp` extension (or re-encode as JPEG if `.jpg` is intentional) so the extension matches the actual format.

README.md
```diff
@@ -75,6 +75,30 @@ deeply analyse complex Django types. Run `mypy` from the root of the web directo
 order to see any type mismatches or errors. The presence of any warning or error will
 immediately fail the PR workflow as well.
 
+## Code Cleanliness
```
zgypa-agent: openai:gpt-5.5 The root `README.md` now has the entire "Code Cleanliness" section duplicated verbatim (currently lines 54-76 and again at 78-99), and this PR adds both copies — likely a merge artifact. Both copies still tell contributors that basedpyright restrictions live in `pyrightconfig.json`, but this PR deletes `pyrightconfig.json` (config appears to have moved into `bfd9000_web/pyproject.toml`). Please de-duplicate the section and update the reference to point at wherever the basedpyright config actually lives now.

zgypa-agent (changes requested):
openai:gpt-5.5 First-round review: documented tests pass at the PR tip (`python manage.py test`: 208 tests; `nix flake check --keep-going`: pass), and the PR description is clear. Six findings posted inline, four of them concrete/reproduced bugs that should be fixed before merge (storage handle leak, Box lookup-cache not invalidated on delete/rmdir, thumbnail WebP bytes saved with a `.jpg` extension, duplicated "Code Cleanliness" docs referencing a deleted `pyrightconfig.json`), plus two open questions (storage delete vs. issue #43, and whether `image_url`/`thumbnail_url` still depend on the old interface per the PR’s own ongoing-efforts list). Requesting changes for the reproduced bugs; happy to re-review once addressed.

zgypa-agent:
openai:gpt-5.5 Follow-up note (not part of the formal review): this PR introduces `bfd9000_thumb`, an entirely new top-level Rust subproject (it does not exist on `develop` at all) — a headless WGPU-based STL thumbnail renderer with its own `Cargo.toml`/`Cargo.lock` (1400+ lines of transitive deps), its own `flake.nix`, and a WGSL shader. It's wired into the root flake as the `stl-thumb` binary and added to both the production Docker `deps` and the dev shell, then invoked via `subprocess` from `generate_stl_thumbnail()` in `media_utils.py`.

Functionally this seems in scope (`DigitalRecord` supports 3D file types, and the PR title says "overhaul file support"), but it's a substantial, untested, GPU-rendering component bundled into a storage-focused PR with no mention in the description — worth flagging for visibility even though the feature itself looks legitimate.

One concrete robustness gap while in there: in `bfd9000_thumb/src/renderer.rs`, `render_stl()` indexes into `mesh.vertices[v_idx]` using face-vertex indices taken directly from the uploaded STL file, with no bounds check. A malformed or malicious STL file with an out-of-range vertex index will panic the renderer process. It's contained today (the Python side calls it as a subprocess with `timeout=5` and `check=True`, so a panic surfaces as a caught `CalledProcessError` rather than crashing Django), but any user-uploadable file can currently force an unhandled panic instead of a graceful parse error. There are also no Rust unit tests for this crate at all. Worth hardening before this renderer processes more untrusted uploads.

zgypa:
bfd9000_thumb is sort of OOS here:
- no documentation
- no tests
- name should be more dscriptive, like `bfd9000_stl_preview` or `_stl_2d_preview`, unless you intend to grow it to handle other thumbs as well.

Also, if you said, you couldn't find a lightweight alternative solution, we could make it its own repo, which we can then re-use for other projects as well. Its strictly not bfd9000 related.

At this stage, i guess you could leave it here, and track this in #98

bfd9000_web/dot-env.example
zgypa, line 1: I actually purposely made it non-hidden, to make sure the dev or other individual sees it right away. Although you might argue that all tools devs use actually show hidden files as well. And to make a clear distinction between example and the real thing.

bfd9000_web/docker-compose.yml
```diff
@@ -15,6 +15,8 @@ services:
       interval: 5s
       timeout: 5s
       retries: 5
+    ports:
```
zgypa, line 19: This docker-compse file could be used as example for deployments as well: the user might  think its required to expose postgres to localhost, while you probably exposed it for dev reasons/debugging. You might want to just add a comment here saying its not needed in production.

zgypa (commented):
Personally, i could find any  blockers here. The agent found some stuff, that it deems requiring changes, so take a look at those, and address them as necessary. I do not see any GUI modifications that i would want to eventually comment, but even if there is something that creeps up, we can address it later.

Thanks

bfd9000_web/archive/storage/storage.py
tnichols217: The goal has shifted to providing a full file interface support, so I believe it is fine to have delete be part of the interface, it is also a requirement for django. Also django will not delete files unless we explicitly call delete on the file from django, so this should be a non-issue

bfd9000_web/archive/storage/box.py
tnichols217: resolved in next commit

bfd9000_web/archive/serializers.py
tnichols217: Updated docs as this is not necessary with our current django structure, as the thumbnail() and image() handlers perform this for us.

bfd9000_web/archive/serializers.py
tnichols217: updated to webp

README.md
tnichols217: resolved

bfd9000_web/docker-compose.yml
tnichols217: fixed

tnichols217:
All items addressed, awaiting push access to repo to commit.

bfd9000_web/archive/storage/urihandler.py
zgypa-agent: openai:gpt-5.6-terra 👍

bfd9000_web/archive/serializers.py
zgypa-agent: openai:gpt-5.6-terra 👍

bfd9000_web/archive/storage/storage.py
zgypa-agent: openai:gpt-5.6-terra The implementation rationale makes sense, but the requested safety policy is not documented. `StorageBackend.delete()` and `rmdir()` remain generic operations, while the storage and data-model docs do not distinguish transient staging files from archival originals or state when app-driven deletion is permitted. Please document the updated policy and guardrails before merge so this is not treated as a general-purpose archive deletion API.

bfd9000_web/archive/serializers.py
zgypa-agent: openai:gpt-5.6-terra Thanks for clarifying the intended storage-URI contract. The current implementation returns the stored locator unchanged: for example, a stored `box://0f9b2c1d.webp` value is serialized as `box://0f9b2c1d.webp`. It does not construct an HTTP API URL.

Could you update `bfd9000_web/docs/api_requirements.md` to make that contract explicit? The current Image URLs section still gives `/api/records/123/image/` as its example and says “Frontend constructs full URL using base API URL.” Those statements are ambiguous or inconsistent with storage locators. Please define these fields as internal backend-qualified storage URIs, give a `box://...` example, and clarify that the image/thumbnail handlers are separate download endpoints.

bfd9000_web/dot-env.example
zgypa-agent: openai:gpt-5.6-terra This rename goes against my earlier intent: `dot-env.example` was deliberately visible, so developers notice it and can clearly distinguish the example from the real hidden `.env` file.

Please restore the visible `bfd9000_web/dot-env.example` filename. The current `bfd9000_web/README.md` already instructs `cp dot-env.example .env`; leaving the file as `.env.example` makes that documented command fail.

zgypa-agent:
openai:gpt-5.6-terra This still needs the requested local CAS/Caddy documentation. `bfd9000_web/README.md` directs developers to `http://127.0.0.1:9000` but does not explain the Caddy proxy at `https://localhost:4430`, while `CAS_ROOT_PROXIED_AS` defaults to that HTTPS origin. The environment example also omits `CAS_ENDPOINT` and `PROXIED_URL`.

Please document the direct-vs-proxied workflow, including that CAS callback testing should use the Caddy HTTPS endpoint, and add the relevant variables to the example configuration.

zgypa-agent:
openai:gpt-5.6-terra Moving the renderer to `bolton-brush/STL-Thumb` addresses the scope concern, and its README/build setup now provide the missing project documentation.

The concrete robustness issue remains in the exact revision this pull request pins (`92765c3`): `renderer.rs` still indexes `mesh.vertices[v_idx]` directly while processing indices from the uploaded STL, with no bounds check, and the repository has no test suite. Please fix this in STL-Thumb by returning a parse/render error for an out-of-range index, add a regression test for malformed index data, then update this pull request flake lock to the fixed revision. This test need not compare rendered pixels.

zgypa-agent:
openai:gpt-5.6-terra The persisted `DigitalRecord.file_size` field and Box item-lookup cache address the record-page metadata overhead, which is good. They do not cache remote file contents, so repeated image/download requests still fetch and stream the object from Box.

Per the earlier discussion, please create and link a separate performance issue for server-side remote-file caching. It should define the cache scope, invalidation policy, size/TTL limits, and whether cached artifacts may contain protected data. This is follow-up work rather than a blocker for this PR.

zgypa-agent:
openai:gpt-5.6-terra 👍 STL thumbnail generation is now implemented through `stl-thumb` and produces a WebP preview. The renderer currently uses a fixed orientation; configurable orientation can be considered separately if a clinical/UI requirement emerges.

zgypa-agent (changes requested):
openai:gpt-5.6-terra Consolidating the outstanding follow-ups here for visibility, since several were raised inline and can be easy to miss in the GitHub UI:

- Document storage deletion guardrails: distinguish transient staging files from archival originals, and define when app-driven `delete()`/`rmdir()` is permitted.
- Finish documenting `image_url` and `thumbnail_url` as internal backend-qualified storage URIs, including a `box://...` example. Remove or clarify the conflicting API-path and “Frontend constructs full URL” guidance.
- Restore the visible `bfd9000_web/dot-env.example` filename, as previously requested. The README currently instructs `cp dot-env.example .env`, which fails while the file is instead named `.env.example`.
- Document the local CAS/Caddy workflow: explain direct port 9000 versus the Caddy HTTPS endpoint at `https://localhost:4430`, and add `CAS_ENDPOINT` and `PROXIED_URL` to the example configuration.
- Fix the unchecked STL vertex-index access and add a malformed-input regression test in `bolton-brush/STL-Thumb`, then update this pull request locked STL-Thumb revision.
- Create and link a separate follow-up issue for server-side remote-file content caching. The current file-size persistence and Box metadata cache improve listing performance, but repeated downloads still stream the object from Box.

The previously reported URI session cleanup, Box lookup-cache invalidation, WebP filename extension, root README cleanup, Docker Compose dev-port note, and STL thumbnail support have been verified and acknowledged.

zgypa (changes requested):
It seems like you might have missed some comments, the github UI is not the best and makes it easy to miss. I had the agent highlight them in a summary comment.

bfd9000_web/dot-env.example
tnichols217: resolved

bfd9000_web/archive/serializers.py
tnichols217: updated

bfd9000_web/archive/storage/storage.py
tnichols217: added new doc

tnichols217:
As for the STL vertex possible out of bound reference, this will just crash the renderer, allowing the python caller to default back to a no-thumbnail fallback. This is fine behavior for now, but can be addressed as an issue on the new STL-Thumb repo instead of as part of this PR. non blocking.

All other items addressed in 64ab20a

tnichols217:
Amended final commit to fix formatting issues, make sure to fetch origin and `reset --hard origin/feature/102-box-support` if youre trying to pull

bfd9000_web/dot-env.example
zgypa-agent: openai:gpt-5.6-terra 👍

bfd9000_web/archive/storage/box.py
zgypa-agent: openai:gpt-5.6-terra 👍

bfd9000_web/archive/storage/storage.py
zgypa-agent: openai:gpt-5.6-terra Addressed directly in `9034b37`. The storage documentation now explains why `delete()` and `rmdir()` remain required backend capabilities while establishing that they are not an archive-authorization boundary. It defines permitted staging cleanup, prohibits generic cleanup/overwrite/synchronization code from deleting archived originals, and requires explicit application authorization for archival deletion. The `source_file` lifecycle documentation now also distinguishes the staging locator from the persistent archival URI, and the backend method docstrings point callers to this policy. This closes the requested guardrail documentation gap. 👍

bfd9000_web/archive/serializers.py
zgypa-agent: openai:gpt-5.6-terra Addressed directly in `4b68236`. The API requirements now define these fields as backend-qualified storage URIs with explicit `<backend-scheme>://<backend-locator>` semantics and a concrete `box://` example. They state that clients treat the values as opaque metadata and retrieve content through the separate authenticated image/thumbnail endpoints. The conflicting “Frontend constructs full URL” and API-path-as-field-value guidance was removed, and implementation notes plus serializer/storage-adapter docstrings now use the same contract. This closes the documentation mismatch without changing the intentional storage-URI implementation. 👍

zgypa (approved):
I had to have the agent intervene with the requested documentation changes, as they were not being addressed as expected.

Issue #52, opened 2026-03-01
Move AI Logic To Backend


zgypa:
Currently its in the front-end. Which means we need to expose BFD9020, which would require authentication and a whole lot of overhead. Just moving the logic to the backend, would allow us to keep BFD9020 inside the docker network, without exposure.

tnichols217:
Fixed in #100

Issue #37, opened 2026-02-27
Secure BFD9020 API access using Docker internal networking


zgypa:
## Problem
The BFD9020 AI classification API should only be accessible from BFD9000 web application, not from external networks.

## Proposed Solution
Use Docker Compose internal networking to isolate the BFD9020 API:

1. Place both BFD9000 web app and BFD9020 API in the same Docker Compose network
2. Do not expose BFD9020 ports to the host (remove port mappings from docker-compose.yml)
3. Configure BFD9000 to access BFD9020 via internal Docker DNS (e.g., http://bfd9020:8000)
4. Only expose BFD9000 web app ports to external network

## Example docker-compose.yml structure
```yaml
services:
  bfd9000_web:
    ports:
      - "8000:8000"
    networks:
      - internal
    environment:
      BFD9020_API_URL: http://bfd9020:8000

  bfd9020:
    # No ports exposed to host
    networks:
      - internal

networks:
  internal:
    driver: bridge
```

## Benefits
- No authentication needed between services
- Simpler configuration
- Better security posture
- Network-level isolation

## Related
- Depends on BFD9020 being containerized
- Related to #22 (Docker deployment)

zgypa:
NB: currently, frontend is calling BFD9020, so that should be moved to backend.

zgypa:
openai/gpt-5.2-codex

Proposing the addition of server-side proxy endpoints for AI calls, as browser-direct requests to internal-only BFD9020 (e.g., at BFD9020_BASE_URL) are failing. Suggest creating Django endpoints  and  that forward requests to BFD9020. This keeps BFD9020 internal-only and aligns with security requirements.

tnichols217:
FIxed in v0.2.1

PR #100, opened 2026-07-23
Fix CI builds, integrate 9020 flake, move AI queries to django backend

tnichols217:
A few updates in this PR:

A lot of work was done to synchronize with the 9020 flake, ensuring that the API is type safe across the services. The current setup uses an auto-generated client package for the 9020 API from the 9020 flake, which is a new dependency of 9000. This tightly couples the API spec from 9020 to 9000, allowing us to notice when there are breaking changes in 9020 in 9000 immediately, as tests would fail. The frontend methods working on the AI classification have been updated to work with the new API spec for 9020. This update allows us to login-protect 9020 and not need to expose 9020 to anything outside the docker network.

CI builds have been updated to as me and Toni have discussed, with hash, staging, and production builds. for easier deployment on wingate. 

A new formatter has also been introduced, in line with the CI development of this PR. The formatting of the django template html files is now also checked and part of the standard formatter with djlint. Accordingly, all template files have been renamed to .dj.html instead of .html in order for the formatters to differentiate between standard html and django templates.

For non-nix users, I've also provided a method to add the 9020 dependency manually, as the development environment normally would place the dependency for you, into the .dummy_deps directory within bfd9000_web. This is gitignored as it would change with the dependencies and there is no reason to vendor it, since we also maintain 9020.

All API routes now are under its own namespace as to reduce collisions with non-API routes, as 9020 is now proxied as an API route as well. This is set up in the new api directory within the archive app.

The docker-compose has also accordingly been updated for the new 9020 config.

Some documentation is likely still lacking for these changes, so that needs to be done before merging, let me know if theres anything underdocumented thus far though.


.github/workflows/publish-ghcr.yml
```diff
@@ -40,24 +39,66 @@ jobs:
           registry: ${{ env.REGISTRY }}
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Fetch all git tags
+        run: git fetch --prune --unimportances --tags
```
zgypa-agent: openai:gpt-5.6-sol: `git fetch` does not support `--unimportances`; it exits immediately with `error: unknown option 'unimportances'`. Because every branch and tag publish reaches this step before metadata extraction and image building, the `build-and-push` job cannot publish any image. Please replace this with the intended valid command, for example `git fetch --prune --tags`.

tnichols217:
Fixes #52 , #37 , standardizes our API and introduces our new linter

bfd9000_web/archive/templates/archive/scan.dj.html
```diff
@@ -1409,27 +1404,24 @@ <h3 class="font-bold text-lg mb-2">Scanner Device Info</h3>
...
-            } else if (result.type_prediction) {
-                const option = state.valuesets.record_types.find(x => x.display === result.type_prediction);
-                if (option) {
-                    recordTypeSelect.value = option.id;
-                }
             } else {
                 showAIStatus('AI classification failed: no valid code returned. Please classify manually.', true);
             }
 
+            const additionalInfo = result.additional_info.prediction
```
zgypa-agent: openai:gpt-5.6-sol: This now reads the correct nested orientation for frontal/lateral images, but `additional_info` is explicitly `null` for every other image class in the locked BFD9020 implementation. Accessing `.prediction` before the null check therefore throws for valid hand, knee, foot, chest, elbow, and hip results. Please keep the nullable result separate, for example `const orientationResult = result.additional_info; const additionalInfo = orientationResult?.prediction ?? null;`, and read rotation confidence from `orientationResult?.probability`.

.github/workflows/publish-ghcr.yml
```diff
@@ -1,18 +1,17 @@
 name: Publish Docker image to GHCR
-# type: line; This workflow intentionally only builds+pushes on semver tags for releases, to avoid multiple image artifacts and ensure clarity.
-
 on:
-  # Only run on v* semver tags (release publish)
   push:
+    branches:
+      - main # Triggers for staging builds
+      - dev # Triggers for dev branch builds
```
zgypa-agent: openai:gpt-5.6-sol: This repository has a `develop` branch but no `dev` branch, and this PR itself targets `develop`. As written, normal development merges never trigger this workflow and the `dev` image tag path is unreachable. Please change both this trigger and the metadata enable condition to `develop`, unless creating and adopting a new `dev` branch is intentionally part of this deployment change.

bfd9000_web/dot-env.example
```diff
@@ -29,7 +29,7 @@ SCANNER_API_BASE=http://localhost:5000
 SCANNER_DEVICE_ID=scanner-001
 
 # BFD9020 AI Service
-BFD9020_BASE_URL=https://wingate.case.edu/bfd9020
+BFD9020_BASE_URL=https://bfd9020:9020
```
zgypa-agent: openai:gpt-5.6-sol: The bundled BFD9020 service listens over plain HTTP, and `docker-compose.yml` correctly configures `http://bfd9020:9020`. Using `https://bfd9020:9020` here causes a TLS handshake against the HTTP Uvicorn server, so the classifier proxy is unavailable for users following this example. `verify_ssl=False` does not correct a protocol mismatch. Please change this value and the root README default to `http://bfd9020:9020` (or document a different reachable URL for host-run development).

zgypa-agent:
openai:gpt-5.6-sol: The four new scan proxy endpoints and their frontend response contract currently have no automated coverage; the existing 208 tests pass without exercising this module. That gap allowed both the original nested-orientation bug and the follow-up null case to survive. Please add focused tests that mock the generated SDK and cover authentication/method enforcement, missing images, successful serialization, upstream validation/errors, and representative `xray-info` payloads with both populated and null `additional_info`.

bfd9000_web/archive/api/scan/views.py
```diff
@@ -0,0 +1,255 @@
...
+        if not uploaded_file or uploaded_file.file is None:  # pyright: ignore[reportUnknownMemberType]
+            return JsonResponse(
+                {
+                    "error": "Missing image file in request. "
+                    + "Expected field name 'image'."
+                },
+                status=400,
+            )
+
+        file_bytes: bytes = uploaded_file.read()  # pyright: ignore[reportAny]
```
zgypa-agent, line 142: openai:gpt-5.6-sol: This reads the entire upload into the Django worker before BFD9020 can enforce its 50 MB `MAX_IMAGE_SIZE`. Django may spool large uploads to disk, but `.read()` still materializes the full file here, so a direct authenticated request can consume unbounded worker memory even though the upstream service will eventually reject it. Please reject `uploaded_file.size` above the classifier limit before reading the payload, and cover that case in the proxy tests.

bfd9000_web/docker-compose.yml
```diff
@@ -77,17 +75,17 @@ services:
     depends_on:
       - bfd9000
   bfd9020:
-    image: ghcr.io/open-ortho/edu.case.bfd9020:local
-    container_name: bfd9020
+    image: localhost/bfd9020:build
     ports:
```
zgypa-agent, line 81: openai:gpt-5.6-sol: Linked issue #37 explicitly requires removing BFD9020 host port mappings so the unauthenticated service is reachable only from BFD9000 over the Compose network. This still publishes port 9020 to host loopback. Loopback prevents remote-machine access, but any browser or process on the host can still call BFD9020 directly, so the requested isolation is not achieved. Please remove `ports` and keep BFD9000 on `http://bfd9020:9020`; if direct local diagnostics are needed, put the mapping in a development override or profile.

tnichols217:
Adding Django tests for the proxy would have had no effect on the status of the other bugs mentioned; we have no testing framework currently for client JS, which was where that bug originated. The generated client library from the 9020 repo already ensures type safety between the two projects in the proxy. Adding full-flow tests from the 9000 side in order to test the proxy with the generated client library would either require reimplementing a mock 9020 for 9000 testing, or spinning up a full 9020 release instance in order to test it.

I think testing the proxy is out of the scope of this PR, since if we were, we really should be creating a client JS testing framework instead with type checkers, linters, and formatters, which is there the found bug would have fallen instead. Additionally, as mentioned in a private message to Toni, we should also create a bruno test suite for full-flow docker-compose tests, which would have also caught the aforementioned bug.

All mentioned areas of testing are irrlevant as they should all instead part of the 9020 test suite, except for authentication verification, which I can add.

If client JS testing and/or a bruno test suite is wanted, we should open a separate issue to track that and leave it out of scope for this PR. Buggy client JS needs to be cleaned up, but is a large task that albeit is related to this PR, is a significantly larger undertaking than the scope of this PR.

tnichols217:
Relevant tests added for requiring image, login, and POST.

zgypa-agent (changes requested):
openai:gpt-5.6-sol: Requesting changes for the unresolved current-head findings.

The publish workflow still fails before building because of the invalid `git fetch --unimportances` option. The documented BFD9020 URL uses HTTPS against the plain-HTTP service. Linked issue #37 still exposes the unauthenticated classifier to the host instead of limiting access to the Compose network. The proxy also buffers uploads before the upstream size limit and lacks integration coverage for its new response and error paths.

There are also three architectural questions and concerns:

- Is there a reason these new `/api/` routes use classic Django views rather than the DRF stack used by the rest of this namespace? CAS establishes a Django session, so DRF `SessionAuthentication` with `IsAuthenticated` should normally preserve the existing SSO behavior while also providing multipart parsing, API exception handling, consistent responses, and schema integration. Would DRF be more convenient and consistent here? If the HTML redirect behavior from `login_required` is intentionally required for these browser-only routes, please document that exception.
- The proxy returns `{"error": "...", "details": "..."}`, while the API specification says all errors use the nested `{"error": {"code": "...", "message": "...", "details": ...}}` contract. Recording the proxy shape in implementation notes creates an exception rather than preserving the API convention; `_handle_client_call()` is the appropriate centralized place to normalize it.
- Extracting API routing from the HTML URL configuration is a good improvement, but `archive/api/api.py` and `archive/api/scan/scan.py` with custom `urls` and `scan_patterns` names introduce a repository-specific routing convention. Conventional `archive/api/urls.py` and `archive/api/scan/urls.py` modules exposing `urlpatterns` would give future development the same organization without requiring special knowledge.

The `develop` trigger and nullable orientation handling are fixed on the latest head and are not blockers. Current CI and the local Django, type, lint, and formatting checks pass, but they do not exercise the publish job or proxy integration.

.github/workflows/publish-ghcr.yml
tnichols217: fixed in next commit

bfd9000_web/archive/templates/archive/scan.dj.html
tnichols217: fixed 2 commits ago

.github/workflows/publish-ghcr.yml
tnichols217: fixed 2 commits ago

bfd9000_web/dot-env.example
tnichols217: fixed in next commit

bfd9000_web/docker-compose.yml
tnichols217: This is the development docker-compose, it should not be used in production. I can add a comment saying not to do this in production, as it is useful in development to be able to directly use those routes.

bfd9000_web/archive/api/scan/views.py
tnichols217: Fixed in next commit with tests

tnichols217:
All 3 structural questions fixed in the next commit; as for why things were imported the way they were, it was how it was previously done, throughout the project, but that's not an excuse to standardize. Fixed

zgypa-agent:
openai:gpt-5.6-sol: Confirmed at the latest HEAD: the three architectural concerns are addressed. The proxy now uses the DRF authentication/request/response stack, errors follow the documented nested contract, and the API modules use conventional `urls.py`/`urlpatterns` routing. Thank you.

.github/workflows/publish-ghcr.yml
zgypa-agent: openai:gpt-5.6-sol: 👍

bfd9000_web/archive/templates/archive/scan.dj.html
zgypa-agent: openai:gpt-5.6-sol: 👍

.github/workflows/publish-ghcr.yml
zgypa-agent: openai:gpt-5.6-sol: 👍

bfd9000_web/dot-env.example
zgypa-agent: openai:gpt-5.6-sol: 👍

bfd9000_web/archive/api/scan/views.py
zgypa-agent: openai:gpt-5.6-sol: 👍

zgypa-agent:
openai:gpt-5.6-sol: Thanks, the added tests now cover authentication, method enforcement, missing and oversized images, the size boundary, and the normalized error envelope. I agree that Django tests would not have caught the JavaScript orientation bug; client-side contract/type checking should be tracked separately.

To clarify, the remaining tests I am requesting here are not JavaScript tests and do not require a live or reimplemented BFD9020 service. `_handle_client_call()` is Django/Python adapter code owned by BFD9000. It can be tested by mocking the generated SDK function and constructing its response objects to cover successful `parsed.to_dict()` serialization, upstream validation/status propagation, empty or malformed parsed responses, and connection or timeout exceptions. Those tests verify how BFD9000 translates generated-client outcomes into its own API contract.

I am satisfied that the request-boundary coverage added here materially improves this PR, but recommend adding these focused Python adapter tests or explicitly tracking them as follow-up work.

bfd9000_web/docker-compose.yml
zgypa-agent: openai:gpt-5.6-sol: I checked the production deployment in `bolton-brush/wingate.docker-configs`. You are correct that this repository's Compose file is development/integration-test orchestration: the separate production Compose does not publish a BFD9020 host port. Keeping `127.0.0.1:9020:9020` here for local diagnostics is therefore reasonable, and I withdraw that part of the concern.

I would not have flagged the local port mapping if the deployment boundary had been documented clearly in these locations:

1. `nix/bfd9000-docker.nix` should have a module-level comment explaining that it builds the production OCI image but does not deploy it, configure runtime networking, or control BFD9020 port publication. The root `README.md` should provide the supporting architectural explanation.

2. The root `README.md`, under a new Deployment Architecture section, should explain that `.github/workflows/publish-ghcr.yml` publishes the Nix-built image to GHCR but does not deploy services. A short header comment in the workflow could reinforce that distinction.

3. `bfd9000_web/docker-compose.yml` should begin with a prominent comment stating that it is development/integration-test orchestration only, that its loopback mappings exist for local diagnostics, and that it must not be used as a production manifest. `bfd9000_web/README.md` should repeat this in its Local Docker Workflow section.

4. The root `README.md` should identify `bolton-brush/wingate.docker-configs` as the current production orchestration repository. Its own `README.md` should confirm that role. The planned NixOS configuration/flake migration should be mentioned only with a link to its tracking issue once that work is formally tracked.

5. `wingate.docker-configs/bfd9000/README.md` should document that issue #37's security boundary is enforced by production orchestration: BFD9020 must use internal service networking and must not have a host port mapping. The application root README should summarize this invariant, while the production configuration should provide the evidence that it is enforced.

One production follow-up remains: `wingate.docker-configs/bfd9000/dot-env-example` currently sets `BFD9020_BASE_URL=https://wingate.case.edu/bfd9020`. To complete the internal-network design, production should use `http://bfd9020:9020`, and the externally accessible Nginx Proxy Manager route for `/bfd9020` should be removed or verified absent.

A short deployment section in the root README would explain this relationship for non-Nix contributors. Concise comments in `nix/bfd9000-docker.nix` and `bfd9000_web/docker-compose.yml` should also state each file's role.

tnichols217: As for the first two statements, I am confused, nix is physically incapable of deploying, it is a build tool. As for the ghcr publish action, it's already documented that it PUBLISHES not deploys. As for the later 3 statements, I can add more comments to clarify those

tnichols217:
The testing of to_dict in this case is guarantee of the actual generated client, if the purpose is to test that it can be serialized properly, we would need to test it against the real package instead of stubbing it, as that would prove nothing, as everything before the to_dict is already type safe and guaranteed by the generated client. I'm not so sure what you intend to test with this as the test you described would just be testing the stub? Are you trying to verify the flow to rethrow an error should the client pass one?

zgypa-agent:
model:claude-sonnet-4-6

You are right, and I am fully withdrawing the request.

Working through it honestly: `_handle_client_call()` has three branches — exception from the SDK call → 502, non-200/None/HTTPValidationError → 502, success → 200 with `parsed.to_dict()`. The logic in each branch is a direct, single-step mapping. There is no subtle condition, no transformation, no state. Type checking already guarantees that `to_dict()` is safe on the success path. The error branches are immediate fallback returns.

The realistic scenario where a test would catch a regression is: someone changes the 502 return to something else, or removes the exception guard. Either of those breaks would surface the moment BFD9020 is unreachable in any integration environment — no dedicated unit test needed. The five or so unit tests I was asking for (exception case, bad status, None parsed, HTTPValidationError, happy path) would each require mocking `_get_api_client()` and constructing SDK response objects — non-trivial ceremony for code that cannot realistically break silently.

The auth, method, and size-boundary tests that are already merged cover the meaningful boundaries BFD9000 owns. I should have reasoned through the failure scenario before making the request. Withdrawing it now.

bfd9000_web/docker-compose.yml
zgypa-agent: model:claude-sonnet-4-6

Confirmed. The `# Ports are not necessary to expose in production, this is just for dev convenience` comment at the `bfd9020` service is there and is exactly what was needed. Thank you.

You are also right about the first two points: a comment on `bfd9000-docker.nix` clarifying that Nix cannot deploy would be noise, and the workflow is already clearly named and documented as a publish step. Both requests were unnecessary.

The wingate.docker-configs documentation points are out of scope for this PR. No further changes needed here.

zgypa-agent (approved):
model:claude-sonnet-4-6

All blockers from the prior CHANGES_REQUESTED review are resolved:

- `git fetch --unimportances` fixed
- BFD9020 URL corrected to plain HTTP
- Proxy rewritten on the DRF stack with `IsAuthenticated`, multipart parsing, and the nested error contract
- API routing follows conventional `urls.py`/`urlpatterns`
- Upload size limit enforced before read; auth, method, size-boundary, and error-envelope tests added
- `docker-compose.yml` annotated as dev-only

Approving.

zgypa (approved):
I like the new convention for django templates, its a good idea. Just make sure its documented somewhere as a new convention: all django html templates SHALL have the dj.html extension. Maybe in AGENTS.md or in developers guide.

tnichols217:
all items addressed. django templates are also documented now. Approved to merge?

tnichols217:
checks passed, merging

